PR #190, opened 2014-05-14
Fix cURL proxy header miscalculation workaround

gws:
**This should be (but does not need to be) merged after #189**

If a proxy header is detected and a buggy cURL version is found, we currently strip the proxy header and reduce the header size by the same amount. This is problematic because the actual cURL bug is that the proxy header size **was not taken into account in the first place**. Therefore, when the header size is reduced on old versions of cURL, the newly calculated header size is too small, and the subsequent calculations are wrong.

Secondly, the `Google_IO_Curl` version constant for the cURL workaround is not correct, the fix version is `7.30.0` (`0x071E00`) [(search for "corrected proxy header response headers count" in the changelog)](http://curl.haxx.se/changes.html), not `7.31.0` (`0x071F00`).

This PR fixes these related problems with the cURL workaround, and retains backward compatibility with the API that was introduced to solve the problem in the first place.

### DIFF
--- a/src/Google/IO/Abstract.php
+++ b/src/Google/IO/Abstract.php
@@ -69,7 +69,12 @@ abstract class Google_IO_Abstract
   abstract public function getTimeout();
 
   /**
-   * Determine whether "Connection Established" quirk is needed
+   * Test for the presence of a cURL header processing bug
+   *
+   * The cURL bug was present in versions prior to 7.30.0 and caused the header
+   * length to be miscalculated when a "Connection established" header added by
+   * some proxies was present.
+   *
    * @return boolean
    */
   abstract protected function needsQuirk();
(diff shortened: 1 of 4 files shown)


Gemmu:
Unfortunately this doesn't work on RedHat 7.2 where the libcurl version 7.29 has been patched with the proxy-header-length fix from 7.30.

> * Thu Nov 27 2014 Kamil Dudka <kdudka@redhat.com> 7.29.0-20
> ...
> - include response headers added by proxy in CURLINFO_HEADER_SIZE (#1161182)

Sadly on both of out production machines (RHEL 7.1 and 7.2) `curl_version()` returns the same info with or without the patch. So it only works on the older machine without the patch.

Please tell if anyone has any suggestions as to how to make it work on both of the machines. Would upgrade to google-api version 2.0 help? I see that it doesn't contain this class at all anymore. This could mean a lot changes to our own code using the client though.

gws:
@Gemmu The only thing I can think of is adding a configurable flag which forces the workaround **off**, skipping the version check. Then, you could add this configuration variable to machines which have the RHEL patch until RHEL catches up to a cURL version where the fix has been implemented.

Having that flag in place wouldn't be a problem for for cURL versions >= 7.30.0 either, which would lower the chances of misconfiguration on an upgrade that bumped the cURL version.

gws:
@Gemmu I've created a branch in my own repo with an example of the above idea: https://github.com/gws/google-api-php-client/commit/26115279c3df36d69997a441885bf4c7e957024f

You would use it like this:

```php
$client->setClassConfig('Google_IO_Curl', 'disable_proxy_workaround', true);
```

I can't think of a good way to test this, but I can submit a PR if you think this would work for you...

mishan:
:+1: I like this idea. I originally authored the patch to not apply the workaround to cURL versions that had the calculation fixed (since it would then break with those.) I can't really think of anything better -- if RHEL patched this and kept the version # as one that's "broken", then it seems that a configuration flag is a good solution.

Gemmu:
Thank you @gws. I copied your fork and setting the flag works as intended. Would it be possible to see a new release on the packagist.org containing this workaround any time soon? I'm no that familiar with the release schedule of the api.

gws:
@Gemmu Great! I'm glad it works for you.

I'm not familiar with the release schedule either, but I've opened #816 - let's take further discussion there once it passes CI and review.

PR #284, opened 2014-08-11
Remove the timezone set.

### LINKED ISSUE #253: Client constructor changes timezone even if it is explicitly set
The client constructor overrides the timezone for PHP here:
https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L83-L85
While it checks if date.timezone is set as a configuration in php.ini, it doesn't check to see if the timezone has been explicitly overridden with `date_default_timezone_set()`. For unfortunate historical reasons, our application does override the timezone, and we had been using `date_default_timezone_set()` to do so.

Now that I see where the bug is coming from, I can set the timezone in our config or use `ini_set()`, but for other users who might have this issue, this might be worth fixing. Unfortunately, I don't know of a way to tell whether a user has explicitly set their timezone with `date_default_timezone_set()`, or I'd submit a pull request.

I also wonder why the timezone is being overridden at all, and if so, why it is optional. Will having the timezone set in the ini file to something other than UTC break any features of the API client?

ianbarber:
Fixes #253 - as far as we can tell the timezone setting adds no benefit
other than removing a warning, which should be addressed in the system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ http://developers.google.com/api-client-library/php
 
 ## Installation ##
 
-For the latest installation and setup instructions, see [the documentation](https://developers.google.com/api-client-library/php/start/installation)
+For the latest installation and setup instructions, see [the documentation](https://developers.google.com/api-client-library/php/start/installation).
 
 ## Basic Example ##
 See the examples/ directory for examples of the key client features.


vincetg2:
Looks like `Google_Http_CacheParser::isExpired()` uses `strtotime()` and `Google_IO_Abstract->makeRequest()` uses `date()`.
Both `strtotime()` and `date()` can be affected by the value of `date.timezone`, so it seems that those uses of these functions may have relied on it being **UTC**.
Would it be worth changing `date.timezone` temporarily and then restoring the original `date.timezone` before and after the use of `strtotime()` and `date()`?

ianbarber:
Yeah, sounds reasonable. Could you raise an issue with that comment?

PR #393, opened 2014-11-15
Update REST.php

lubberscorrado:
try gzdecode on body if sent in binary format


lubberscorrado:
i still/again have that problem and used the most recent commit. i implemented the patch to make it work again

PR #437, opened 2014-12-20
Fix #400: move autoloader to src/Google

### LINKED ISSUE #400: Move autoloader to Google/ directory
In Fedora, we have the library packaged below `/usr/share/php/Google` . With the old arrangement where the library itself and things that used it would do something like this:

    require_once Google/Utils.php;

that worked fine - as /usr/share/php is on the system include path, anything that used the library would find the system-wide copy if it was installed. However, the new autoload mechanism is hardcoded to expect the autoloader to be two levels *beneath* the Google directory, stuff like:

    require_once realpath(dirname(__FILE__) . '/../../../autoload.php');

That means we'd have to ship it in `/usr/share`, which is obviously not a good design.

Is there any reason the autoloader can't just be in the Google directory? It'd make things easier.

AdamWill:
This makes system-wide installation of the library more viable.

It's a quick job, but it looks fine. I ran the test suite and got the same results before and after.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -15,19 +15,12 @@
  * limitations under the License.
  */
 
-function google_api_php_client_autoload($className) {
-  $classPath = explode('_', $className);
-  if ($classPath[0] != 'Google') {
-    return;
-  }
-  if (count($classPath) > 3) {
-    // Maximum class file path depth in this project is 3.
-    $classPath = array_slice($classPath, 0, 3);
-  }
-  $filePath = dirname(__FILE__) . '/src/' . implode('/', $classPath) . '.php';
-  if (file_exists($filePath)) {
-    require_once($filePath);
-  }
+// PHP 5.2 compatibility: E_USER_DEPRECATED was added in 5.3
+if (!defined('E_USER_DEPRECATED')) {
+  define('E_USER_DEPRECATED', E_USER_WARNING);
 }
 
-spl_autoload_register('google_api_php_client_autoload');
+$error = "google-api-php-client's autoloader was moved to src/Google/autoload.php in 1.1.3. This ";
+$error .= "redirect will be removed in 1.2. Please adjust your code to use the new location.";
+trigger_error($error, E_USER_DEPRECATED);
+require_once 'src/Google/autoload.php';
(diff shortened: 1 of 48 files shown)


AdamWill:
"On the off chance I had a check, but didn't get a hit on owncloud on the corp signers list. "

I don't work for ownCloud, I work for Red Hat (I package ownCloud for Fedora as a side project). I'm trying to find out if RH legal has an opinion on the individual CLA and/or we have any corporate agreement ATM.

ianbarber:
Sounds good!

autoload.php
```diff
@@ -15,19 +15,12 @@
...
+// PHP 5.2 compatibility: E_USER_DEPRECATED was added in 5.3
+if (!defined('E_USER_DEPRECATED')) {
+  define('E_USER_DEPRECATED', E_USER_WARNING);
 }
 
-spl_autoload_register('google_api_php_client_autoload');
+$error = "google-api-php-client's autoloader was moved to src/Google/autoload.php in 1.1.3. This ";
+$error .= "redirect will be removed in 1.2. Please adjust your code to use the new location.";
+trigger_error($error, E_USER_DEPRECATED);
+require_once 'src/Google/autoload.php';
```
glensc: i suggest not  to rely on relative paths, as those get resolved from include_path. 

use sth like this instead:
```php
require_once  dirname(__FILE__) . '/src/Google/autoload.php';
```

glensc:
how about dropping "require_once" lines from classes? the documentation already says you should first load autoloader, meaning you do not need to do that again from classes themselves.

AdamWill:
@glensc AIUI, the `require_once` are there for backwards compatibility, to make existing projects (which directly require the files they need, as was previously the only option) keep working with the new version.

autoload.php
AdamWill: good point, thanks.

ianbarber:
Any joy on the CLA?

AdamWill:
Sorry, still didn't hear back from the right person - I'm gonna give it a day or two then just go ahead and sign the personla version.

AdamWill:
I signed the individual CLA today, sorry for the delay!

ianbarber:
Thanks!

PR #447, opened 2015-01-12
autoload.php generates library conflicts in CMSs

deconf:
CMS extension and/or plugin developers should customize autoload.php to avoid "redeclare fatal errors"

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -15,7 +15,10 @@
  * limitations under the License.
  */
 
-function google_api_php_client_autoload($className) {
+/**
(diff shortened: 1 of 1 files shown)


ianbarber:
Thanks! I'm not sure that renaming the function is super worthwhile though - perhaps its easier to add a note to the README with your recommendation?

deconf:
We could also check the function name, and if it wasn't customized we could throw an exception. But that may be a drawback for those using the library outside a CMS.

A recommendation in the README it's a good idea.

ianbarber:
Added a note to the README, thanks!

PR #460, opened 2015-01-22
Update Resource.php

Kralock:
We can't send a zip file, cause zip file is not a media.


ianbarber:
Media should work fine for the zip files - I'm not sure what this patch is trying to do?

PR #461, opened 2015-01-22
Allow cURL options to be set through setClassConfig

deconf:
Will allow developers to set additional cURL options when needed.

For example, googleapis.com resolves to both IPv4 and IPv6 addresses now. Some hosting providers don't have IPv6 support and a cURL request will fail when trying to connect to an IPv6 address.

Being able to set cURL options, developers can handle this issue using CURLOPT_IPRESOLVE to force IPv4, IPv6 or Both.

### DIFF
--- a/src/Google/IO/Curl.php
+++ b/src/Google/IO/Curl.php
@@ -80,7 +80,12 @@ class Google_IO_Curl extends Google_IO_Abstract
     if ($request->canGzip()) {
       curl_setopt($curl, CURLOPT_ENCODING, 'gzip,deflate');
     }
-
+    
+    $options = $this->client->getClassConfig('Google_IO_Curl', 'options');
+        if (is_array($options)){
+        $this->setOptions($options);
+    }
+    
     foreach ($this->options as $key => $var) {
       curl_setopt($curl, $key, $var);
     }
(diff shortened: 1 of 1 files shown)


src/Google/IO/Curl.php
```diff
@@ -80,7 +80,12 @@ public function executeRequest(Google_Http_Request $request)
     if ($request->canGzip()) {
       curl_setopt($curl, CURLOPT_ENCODING, 'gzip,deflate');
     }
-
+    
+    $options = $this->client->getClassConfig('Google_IO_Curl', 'options');
+        if (is_array($options)){
```
ianbarber: tabbing is off here, and can we have a space before the opening { Thanks!

ianbarber:
Thanks! Have you had a chance to sign the CLA? See https://github.com/google/google-api-php-client/blob/master/CONTRIBUTING.md

deconf:
I've signed it few minutes ago, thanks!

ianbarber:
Thanks!

Kak3n:
Hello could I ask how exactly I can set cURL options?

bshaffer:
@Kak3n if you are using `v2`, you would set them through the guzzle library. See [this page](http://docs.guzzlephp.org/en/latest/request-options.html) for more info.

Kak3n:
thanks btw what is the difference betwen v1 and v2?

Dňa streda, 25. mája 2016 Brent Shaffer <notifications@github.com>
napísal(a):

bshaffer:
`v2`  is the current version, but the docs on `developer.google.com` have not been updated for them yet. If you are just starting your project, use v2

Kak3n:
thanks.

2016-05-25 20:51 GMT+02:00 Brent Shaffer <notifications@github.com>:

PR #474, opened 2015-02-05
added fusion table example

dan-klasson:


ianbarber:
Thanks! We don't generally include service specific examples, just ones that demonstrate generic features of the client library. I'd recommend posting this as a gist , and note of the version of the library it was built against.

PR #475, opened 2015-02-05
flock() not getting file handle all the time

kylefarris:
Removed the assumption of successful file handle acquisition while trying obtain a lock on a cache file. This has caused the following errors to pop up in my logs:

`flock() expects parameter 1 to be resource, boolean given`

In other words we shouldn't try and get a lock on a file we can't open. I suppose the eventual solution would be to find why the file can't be opened in the first place, but, this only happens occasionally so it might be some kind of race condition or something.


ianbarber:
Looks good, thanks!

PR #478, opened 2015-02-06
Added support for PHP 5.2.x for the Google P12 class sign method...

mxaddict:


quantumpacket:
Signing data which is security related would probably not be a concern for anyone who has not updated from PHP 5.2. There are reasons for minimum version requirements. PHP 5.2 had reached EOL in January 2011.

Anyone still running that version should choose a different profession, as they have been running a vulnerable version of PHP for over 4yrs. I don't think we should be encouraging the use of such outdated and insecure versions by adding support for it.

:-1:  -1 on this change.

mxaddict:
Anyone still running that version should choose a different profession, as they have been running a vulnerable version of PHP for over 4yrs. I don't think we should be encouraging the use of such outdated and insecure versions by adding support for it.

<<<< So that is the reason why 5.2.1 is the officially listed version that the Library supports? on README?

mxaddict:
BTW, we are running a Legacy app on internal servers with no external access with over 3K files that would all need editing to support any version of PHP that is newer than 5.2.x.

quantumpacket:
Minimum version requirements is always a hot topic :P The readme states why they chose that minimum version. If we are to prevent end users from shooting themselves in the foot by using vulnerable versions, we should stop handing them the bullets. Each new lower version supported just adds more code to maintain for backwards compatibility, and says to the end user "hey, go ahead and use a vulnerable version because we support you on that".

With PHP7 on the horizon, which would have PHP 5.4 reach EOL shortly after, do we really want to be adding additional support that many EOLed versions back? I'd be fine for adding additional support for maybe 5.3, 5.4 even though they are EOL, but there has to be a point where it's just not advised. If the PHP community is to ever upgrade from 5.2, then major vendors needs to stop supporting such old versions.

Your legacy app can be more relaxed with updating, due to it being on an internal server. This repo on the other hand is likely to be used on primarily public facing services, which should not be encouraged to using vulnerable versions.

I don't want to turn this into a flame war about version requirements. So not going to argue about it. I apologize if I came off as rude or anything, just stating my opinion on adding additional support for such old versions.

eRadical:
A "recommended" section should be introduced but the library should not «force» a version for education purpose. The "minimum" is as it says «minimum» = if you are going to use it with a previous version than the "minimum" the library is not going to be functional. Functionality is already crippled under 5.2.x.

A library purpose is to offer functionality not enforce a version.
YES... we definitely need to put pressure on DevOps and/or system administrators to upgrade PHP version... but that is completely another thing.

[Please don't get me wrong I personally use 5.5 and do not have any server with PHP under 5.4... but I make it my business to be interested in this.]

mxaddict:
I think this thread is going to start a flame war, So I will be closing it :100: %

Issue #411, opened 2014-11-26
Patch a NULL value (for example: All-day event to Timed event)


mail2rvg:
Hi there,

We're experiencing an issue when we want to patch the (start/end) Date/DateTime properties of an all-day-event. We want to change it from an all-day event to a timed event.

We create a new instance of Google_Service_Calendar_EventDateTime and set the desired DateTime value using the setDateTime method, next we set the Date value to NULL using the setDate method. When patching this event we got this error: 'Invalid or mismatching start and end times'.

This is due to the fact the date property isn't present in the request body. The toSimpleObject method in Google_Model class processes all the public properties, but skips the properties that has a NULL value. This would mean we could never patch a property to a NULL value.

We tried to disable the NULL value check, so the date property with the NULL value will be placed in the request body. This led to the desired result; the all-day event in Google changed to a timed event.

We were wondering if the patch semantics prevent patching a null value, so we read the rfc
http://tools.ietf.org/html/rfc5789. It does not seem the case, so this is why we filled this bug report.

Following https://developers.google.com/google-apps/calendar/v3/reference/events the date property is only used for a all-day event.

Thanks in advance

alexkvak:
+1

yapro:
+1

TheXardas:
+1

denlapaev:
+1

Strate:
+

dweingart:
This code (stripping null attributes) also causes problems for the Directory API.

If you want to clear any of the array attributes of the Google_Service_Directory_User object (like emails, phones, or addresses) setting these properties to the empty array doesn't work. Only sending NULL (in the JSON) works to clear the attribute in my testing.

However, there is no way to set the properties to NULL and have the nulls preserved with the current Google_Model code.

ianbarber:
Because the template generated files have properties and getter/setters I couldn't see a sensible way to automatically tracked uninitialised variables, so as a workaround I've added a Google_Model::NULL_VALUE const. If you set a value to that, it will be preserved as null in the JSON. If anyone has better/other ideas would love to see a PR on it though!

viralpandya:
Having same issue (c#) and not while sending null value(NULL_VALUE) with Datetime, it gives an error of malformat of date value.

Anyone to help?

bshaffer:
@viralpandya have you tried using the `Google_Model::NULL_VALUE` constant instead of PHP `NULL`? 

[...]

PR #481, opened 2015-02-13
Patch a NULL value (All-day event to Timed event)

yapro:
Patch a NULL value (for example: All-day event to Timed event) - https://github.com/google/google-api-php-client/issues/411


ianbarber:
I'm not sure about the $this->id property check - I can see the check on modelData, but I'm not sure this is addressing the general case? I may be misunderstanding the code though, so please let me know your thoughts.

yapro:
I prepared the new patch https://github.com/google/google-api-php-client/pull/484

Issue #480, opened 2015-02-12
Downloading file/object from google cloud storage


pulkitjalan:
I can't seem to figure out how to download an object from google cloud storage using this client. I have got as far as getting the `Google_Service_Storage_StorageObject ` object which contains the information about the object, but what next?

Thanks in advance for your help!

ianbarber:
From a quick look it looks like it could be using alt=media as a parameter: https://cloud.google.com/storage/docs/json_api/v1/objects/get . You may run into trouble with the lib trying to parse the response as JSON though, so do give it a go If that's not it maybe try asking the cloud support team how the API is supposed to be used? https://cloud.google.com/storage/docs/resources-support#community 

Alt media download might be a useful example though, so if they give an answer please consider updating this ticket!

pulkitjalan:
Hi @ianbarber 

Thanks for your reply, I tried this out however ran into the problem where the library is trying to parse the response as json...

Below is what I tried, and as far as I can tell it is downloading the file.

```php
$file = $storage->objects->get($bucket, $path, ['alt' => 'media']);
```

When running this I get a `Google_Service_Exception` with the message `Invalid json in service response`

Its caused in `src/Google/Http/REST.php` on line `123`.

```php
if ($code != '204') {
    $decoded = json_decode($body, true);
    if ($decoded === null || $decoded === "") {
        $error = "Invalid json in service response: $body";
        if ($client) {
            $client->getLogger()->error($error);
        }
        throw new Google_Service_Exception($error); // <=========== HERE
    }

    if ($response->getExpectedClass()) {
        $class = $response->getExpectedClass();
        $decoded = new $class($decoded);
    }
}
```

The response code for this request was `200`, so could it be as simple as adding another condition in this file?

dheeraj143:
@ianbarber @pulkitjalan Thank you guys! This helped me: 
`$file = $storage->objects->get($bucket, $object, ['alt' => 'media']);` 
:+1:

shyandsy:
@dheeraj2dj there is no $storage->objects now

PR #483, opened 2015-02-14
Fix for file download from cloud storage

pulkitjalan:
Currently when running a get request on storage with `alt=media` the library will try and decode the response into json, however when `alt=media` is specified to download the file the api returns the file contents as the body. In this case the library should simply return the raw body without trying to decode. #480

### DIFF
--- a/src/Google/Http/REST.php
+++ b/src/Google/Http/REST.php
@@ -113,6 +113,10 @@ class Google_Http_REST
 
     // Only attempt to decode the response, if the response code wasn't (204) 'no content'
     if ($code != '204') {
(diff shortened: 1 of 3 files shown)


ianbarber:
Can you drop the change to the Storage class - that's generated from a Discovery service description so it wont be persisted unfortunately. Otherwise, looks great! Just to check - have you signed the CLA: https://github.com/google/google-api-php-client/blob/master/CONTRIBUTING.md

pulkitjalan:
Done, yea just signed the CLA. Thanks.

ianbarber:
Thanks!

PR #484, opened 2015-02-16
Patch a NULL value (All-day event to Timed event)

yapro:
Permit the use of "null" in the second level of nesting.
Details - https://github.com/google/google-api-php-client/issues/411

### DIFF
--- a/src/Google/Model.php
+++ b/src/Google/Model.php
@@ -142,7 +142,7 @@ class Google_Model implements ArrayAccess
     foreach ($props as $member) {
       $name = $member->getName();
       $result = $this->getSimpleValue($this->$name);
-      if ($result !== null) {
+      if (!property_exists($this, 'id') || $result !== null) {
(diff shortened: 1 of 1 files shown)


alexkvak:
+1

ianbarber:
The this ->id check seems odd - was that suppose to be looking for a named property in specific, or in general?

yapro:
this ->id using in the first level of nesting (it checking looking for the named property in specific nesting).

yapro:
I signed CLA on the page https://cla.developers.google.com/clas

ianbarber:
I wanted a more generalised fix, so I've added a Google_Model::NULL_VALUE const you can use to pass which will be preserved as null. I'll close this one, but if you have an idea for tracking uninitialised variables that works within the current service class structure, would love to see it. Thanks!

yapro:
Your idea is good. I will wait implementation.

PR #492, opened 2015-02-24
avoid searching include path

glensc:
from #437 comments:

https://github.com/google/google-api-php-client/pull/437#discussion-diff-23418551


ianbarber:
Thanks!

Issue #440, opened 2014-12-21
Make composer use the distributed autoloader


whatthejeff:
Once #437 is completed and merged, it might be nice to update `composer.json` to use the autoloader that is distributed with this project. This would prevent composer from generating an unnecessary classmap and would prevent composer users from having multiple autoloaders for the same classes.

~~~diff
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
       "phpunit/phpunit": "3.7.*"
     },
     "autoload": {
-        "classmap": [
-            "src/"
+        "files": [
+            "src/Google/autoload.php"
         ]
     },
     "extra": {
~~~

glensc:
i prefer composer based autoloader, because composer can write optimized version (-o). and with composer there will be only one spl_autoload method registered, instead of two (composer one and this one from project)

when using composer src/autoload.php from this project should be ignored.

so -2 for this PR

whatthejeff:
> when using composer src/autoload.php from this project should be ignored.

@glensc I agree with you. The problem is that many files in this project manually include the non-composer autoloader. You end up with both autoloaders loaded regardless. This suggestion was only aimed at making sure you don't have two autoloaders responsible for the same classes.

whatthejeff:
@glensc FWIW, I also prefer composer's optimized autoloaders. I think I even suggested using a classmap as the default autoloader in a different ticket.

glensc:
well, the classloader from classes should be [removed eventually](https://github.com/google/google-api-php-client/pull/437#issuecomment-71117638). but in my own projects i've used such hack:

```php
if (!class_exists('Composer\Autoload\ClassLoader', false)) {
// setup classloader for outside composer usage
}
```

this could be put to this project's `autoload.php` to avoid two of them being setup.

whatthejeff:
@glensc Yeah, I had also considered that solution. My concern was that it could break autoloading for projects that are using composer for some libs, but have used an alternate install method for this lib.

AdamWill:
@whatthejeff that happens to be exactly owncloud's situation, in fact - it uses composer for the 'core' top-level 3rdparty/ directory, but the google lib is included by an 'application' and not managed by composer.

glensc:
ah yes, [my solution](https://github.com/google/google-api-php-client/issues/440#issuecomment-71164196) only solves for Composer autoloader.

glensc:
so, maybe in classes which load `autoload.php` use such snippet:

```php
[...]

PR #493, opened 2015-02-24
setup autoloader from classes only if no autoloader detected

glensc:
see #440 comments for details


ianbarber:
And for this one!

PR #507, opened 2015-03-17
Fixed auth by service account json file

vlastv:


vlastv:
I just submitted the CLA.

ianbarber:
Thanks! I'm a bit concerned about the naming - let me check there aren't any routes where you get files with the old format still. May be that we need to support both.

ianbarber:
Ended up merging #524 which handled both versions, so this functionality should be in. Sorry  for the delay, thanks though!

PR #509, opened 2015-03-18
Change servicePath for file upload with parent

giltotherescue:
The Google API returns a 400 error when files are uploaded to the default service path, but only when a parent folder is specified. The error message says: _Uploads must be sent to the upload URL_. This error can be resolved by prepending ```upload/``` to the service path URL.

### DIFF
--- a/src/Google/Service/Drive.php
+++ b/src/Google/Service/Drive.php
@@ -1500,7 +1500,11 @@ class Google_Service_Drive_Children_Resource extends Google_Service_Resource
   {
     $params = array('folderId' => $folderId, 'postBody' => $postBody);
     $params = array_merge($params, $optParams);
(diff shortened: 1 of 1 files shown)


giltotherescue:
I just submitted the CLA.

giltotherescue:
I checked the build failure. Here is what it says:

```PHP Fatal error:  Cannot redeclare class Google_Service_SQLAdmin_OperationError in /home/travis/build/google/google-api-php-client/src/Google/Service/SQLAdmin.php on line 2740```

I did not make any changes to this file, so there must be a pre-existing bug in this release.

ianbarber:
This is a good idea, but the Drive class is generated so your change would get overwritten unfortunately :(. Could you open an issue with how you are doing the upload at the moment, there may be a fix we can do elsewhere in the library to resolve this more generally.

PR #515, opened 2015-04-02
Example of Calendar.

ransom1538:
* New example of calendar 
* This also shows how Scopes work (which has caused plenty of stack overflow posts).
* I have signed the google doc for pushing/forking code

### DIFF
--- a/examples/calendar.php
+++ b/examples/calendar.php
@@ -1,0 +1,126 @@
+<?php
+session_start();
+/*
+ * Copyright 2015 Google Inc.
+ *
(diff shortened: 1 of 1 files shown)


ianbarber:
Thanks! That's a really great sample, but we don't host API specific samples unless they're demonstrating a core library feature. This is largely because they get out of date as APIs change and we don't have the familiarity with every API to keep them following every specific best practice. We've recommended that teams create their own samples repo in the past, so perhaps creating your own repo and pulling in the library via composer dependency. If you do that, we can add a link to your examples repo from the README here.

Issue #463, opened 2015-01-26
Translation detections too deep in modelData array?


egeriis:
I am currently implementing translation detection in my application. While testing the Google API PHP client I seem to have stumbled upon an issue.

To get a glimpse at what `listDetections` returns, I've `var_dump`ed the return value, which is as such:

```
object(Google_Service_Translate_DetectionsListResponse)#23 (6) {
  ["collection_key":protected]=>
  string(10) "detections"
  ["internal_gapi_mappings":protected]=>
  array(0) {
  }
  ["detectionsType":protected]=>
  string(48) "Google_Service_Translate_DetectionsResourceItems"
  ["detectionsDataType":protected]=>
  string(5) "array"
  ["modelData":protected]=>
  array(1) {
    ["data"]=>
    array(1) {
      ["detections"]=>
      array(1) {
        [0]=>
        array(1) {
          [0]=>
          array(3) {
            ["language"]=>
            string(2) "da"
            ["isReliable"]=>
            bool(false)
            ["confidence"]=>
            float(0.024751695)
          }
        }
      }
    }
  }
  ["processed":protected]=>
  array(0) {
  }
}
```

After looking at this I found that this return type, `Google_Service_Translate_DetectionsListResponse`, includes a `getDetections` method. Calling this however, returns and empty array, even though there is a language detection.

So I looked a bit more into how the PHP lib works, and from the looks of it, it seems that the `modelData` is a hash with one key: "data", which then contains a hash with one key: "detections". This seems odd given that the lib seems to query on `modelData[collection_key]` and `collection_key` is set to "detections" for `Google_Service_Translate_DetectionsListResponse`.

Is this a bug or am I looking at it the wrong way?

ianbarber:
Yeah, this looks odd. Do you have a small script we could use to repro?

egeriis:
This should do?

```php
$client = new Google_Client();
$client->setDeveloperKey('...');
$translate = new Google_Service_Translate($client);
$translations = $translate->translations->listTranslations($this->description, $targetLanguage);
```

davidrojo:
Same here, the problem, you can checkt it with this code:

```php
$client = new \Google_Client();
$client->setApplicationName("AutomatedTranslations");
$client->setDeveloperKey(self::GOOGLE_API_KEY);
        
$service = new \Google_Service_Translate($client);
$result = $service->translations->listTranslations('Hola buenos dias', 'en', ['source' => 'es']);
[...]

PR #523, opened 2015-04-10
Fix empty translation bug - issue #463

tegansnyder:


ianbarber:
Thanks! Unfortunately this service is generated, so you change would get overwritten :( The fix we need to do is to change the generator - it looks like there is a "feature" - "dataWrapper" flag in their to indicate the API is using that approach, and we need to have code to handle that case. Judging by previous experience that is likely to take a while, so I might be open to adding a specific check for Translate API calls in the REST.php decode response, and doing the data unwrapping there. Its ugly, but AFAIK this is the only API with this feature.

tegansnyder:
Makes sense :) thanks for the explanation

PR #524, opened 2015-04-10
Added support for newer json service accounts

nebez:
Left old json service account functionality untouched to ensure
backwards compatibility isn't broken. Added test for both formats of
json service files.


ianbarber:
Looks great, thanks!

Issue #525, opened 2015-04-13
Unused variables in IoTest.php 


kesar:
It's not necessary to pass variable ```$client``` anymore in the test ```IoTest```

PR #526, opened 2015-04-13
Removed unused variables in IoTest.php

kesar:
Fixes https://github.com/google/google-api-php-client/issues/525


ianbarber:
Thanks!

Issue #527, opened 2015-04-13
Unused private variables in Authentification classes


kesar:
Private variable ```$key``` is not setted in classes Google_Auth_AppIdentity & Google_Auth_Simple

PR #528, opened 2015-04-13
Unused private variables in Authentification classes

kesar:
Fixes https://github.com/google/google-api-php-client/issues/527


ianbarber:
Good spot!

Issue #529, opened 2015-04-13
Wrong documentation in Abstract.php


kesar:
Current documentation use ```Google_HttpRequest``` instead of ```Google_Http_Request```

PR #530, opened 2015-04-13
Wrong documentation in Abstract.php

kesar:
Fixes https://github.com/google/google-api-php-client/issues/529


ianbarber:
Thanks!

Issue #531, opened 2015-04-14
Duplicated parameter in Resource.php


kesar:
In Resource.php there is a duplicated parameter userip

![screen shot 2015-04-14 at 20 36 32](https://cloud.githubusercontent.com/assets/1288106/7144503/4d55b2be-e2e6-11e4-8695-ae8eacedd6b1.png)

![screen shot 2015-04-14 at 20 36 56](https://cloud.githubusercontent.com/assets/1288106/7144504/4d5af8d2-e2e6-11e4-8c3b-38a277f0754d.png)

PR #532, opened 2015-04-14
Removed (duplicated) userip parameter

kesar:
Fixes https://github.com/google/google-api-php-client/issues/531


ianbarber:
Must admit, not sure if this was a mistake or an attempt to support either version capitalisation wise. I'm relatively happy with removing it for now, but we might see some BC issues for people that were supplying it formatted differently.

Issue #533, opened 2015-04-14
Recursion problem between Service and Resources


kesar:
Right now its stored into the resources (Google_Service_Resource) the service (Google_Service). This cause a problem of recursion.

![screen shot 2015-04-14 at 21 03 28](https://cloud.githubusercontent.com/assets/1288106/7145047/c1fd12b2-e2e9-11e4-84b2-fb5971eabc3f.png)

PR #534, opened 2015-04-14
Removed service to avoid recursion problem

kesar:
In order to avoid problems with recursion, I would suggest to store in the resource object only the information necessary from the service, in this case path.

Fixes https://github.com/google/google-api-php-client/issues/533


ianbarber:
Looks good.

PR #548, opened 2015-04-27
More efficient file upload from URL

woofyman99:
While the sample code works file from a local file.  If you're reading from a URL, it's very inefficient.  The sample code here is reading from a URL and is very inefficient.

https://developers.google.com/youtube/v3/code_samples/php#resumable_uploads

It looks like it was pulled directly from the sample code in GitHub


ianbarber:
Any chance you could sign in the CLA - link in the bot comment? Understand if not, but we need it for any changes.

woofyman99:
No problem, done.

ianbarber:
Thanks!

PR #549, opened 2015-05-01
Avoid double URL encoding for params' values

deconf:
According to documentation, for filters, URL-reserved characters — Characters such as & must be url-encoded in the usual way, see https://developers.google.com/analytics/devguides/reporting/core/v3/reference#filterExpressions

If in a filter such as ga:pagePath an URI is used, which needs to be url-encoded, the createRequestUri method will apply an additional rawurlencode() on it resulting in a double URL encoding.


ianbarber:
Thanks!

PR #561, opened 2015-05-19
Allow for valid cross client authentication requests

tehmaestro:
Here's a pull request to allow for valid authenticate requests, using an authorization code obtained through the cross client authorization scheme, as described here https://developers.google.com/identity/protocols/CrossClientAuth

  As the documentation sais, the request body should not include the request_uri argument.

### DIFF
--- a/src/Google/Auth/OAuth2.php
+++ b/src/Google/Auth/OAuth2.php
@@ -82,7 +82,7 @@ class Google_Auth_OAuth2 extends Google_Auth_Abstract
    * @throws Google_Auth_Exception
    * @return string
    */
-  public function authenticate($code, $crossClient)
+  public function authenticate($code, $crossClient = false)
   {
     if (strlen($code) == 0) {
       throw new Google_Auth_Exception("Invalid code");
--- a/src/Google/Auth/OAuth2.php
+++ b/src/Google/Auth/OAuth2.php
@@ -78,13 +78,25 @@ class Google_Auth_OAuth2 extends Google_Auth_Abstract
 
   /**
    * @param string $code
+   * @param boolean $crossClient
    * @throws Google_Auth_Exception
    * @return string
    */
-  public function authenticate($code)
+  public function authenticate($code, $crossClient)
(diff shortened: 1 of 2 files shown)


tehmaestro:
I signed the cla.

src/Google/Auth/OAuth2.php
```diff
@@ -78,28 +78,34 @@ public function authenticatedRequest(Google_Http_Request $request)
 
   /**
    * @param string $code
+   * @param boolean $crossClient
    * @throws Google_Auth_Exception
    * @return string
    */
-  public function authenticate($code)
+  public function authenticate($code, $crossClient)
```
kesar: if this is public I would make it optional

src/Google/Client.php
```diff
@@ -114,15 +114,18 @@ public function getLibraryVersion()
...
+   * If $crossClient is set to true, the request body will not include
+   * the request_uri argument
    * Helper wrapped around the OAuth 2.0 implementation.
    *
    * @param $code string code from accounts.google.com
+   * @param $crossIdentity boolean, whether this is a cross-client authentication
    * @return string token
    */
-  public function authenticate($code)
+  public function authenticate($code, $crossClient = false)
```
kesar: I would add some UT related to crossClient

src/Google/Client.php
```diff
@@ -114,15 +114,18 @@ public function getLibraryVersion()
 
   /**
    * Attempt to exchange a code for an valid authentication token.
+   * If $crossClient is set to true, the request body will not include
+   * the request_uri argument
    * Helper wrapped around the OAuth 2.0 implementation.
    *
    * @param $code string code from accounts.google.com
+   * @param $crossIdentity boolean, whether this is a cross-client authentication
```
kesar: param is ```crossClient``` not ```crossIdentity```

kesar:
:+1:

ianbarber:
Reasonable approach, thanks!

PR #562, opened 2015-05-19
Add support for per-service root url definitions

ircmaxell:
This adds support for per-service root url definitions.

### DIFF
--- a/src/Google/Service/Resource.php
+++ b/src/Google/Service/Resource.php
@@ -41,6 +41,9 @@ class Google_Service_Resource
       'prettyPrint' => array('type' => 'string', 'location' => 'query'),
   );
 
+  /** @var string $rootUrl */
+  private $rootUrl;
+
   /** @var Google_Client $client */
   private $client;
 
@@ -58,11 +61,12 @@ class Google_Service_Resource
 
   public function __construct($service, $serviceName, $resourceName, $resource)
   {
+    $this->rootUrl = $service->rootUrl;
     $this->client = $service->getClient();
     $this->servicePath = $service->servicePath;
     $this->serviceName = $serviceName;
     $this->resourceName = $resourceName;
-    $this->methods = isset($resource['methods']) ?
+    $this->methods = is_array($resource) && isset($resource['methods']) ?
         $resource['methods'] :
         array($resourceName => $resource);
   }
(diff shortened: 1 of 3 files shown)


src/Google/Service/Resource.php
```diff
@@ -58,6 +61,7 @@ class Google_Service_Resource
 
   public function __construct($service, $serviceName, $resourceName, $resource)
   {
+  	$this->rootUrl = $service->rootUrl;
```
ianbarber: Fix tab.

tests/general/ResourceTest.php
```diff
@@ -0,0 +1,124 @@
...
+    $this->client->expects($this->any())
+          ->method("getAuth")
+          ->will($this->returnValue(new Google_Auth_Simple($this->client)));
+    $this->service = new Test_Google_Service($this->client);
+  }
+
+  public function testCallFailure()
+  {
+    $resource = new Google_Service_Resource($this->service, "test", "testResource", "testMethod");
+    $this->setExpectedException("Google_Exception", "Unknown function: test->testResource->someothermethod()");
```
ianbarber: line length - can we keep to 100 width.

src/Google/Service/Resource.php
```diff
@@ -193,7 +197,12 @@ public function call($name, $arguments, $expected_class = null)
         null,
         $postBody
     );
-    $httpRequest->setBaseComponent($this->client->getBasePath());
+
+    if ($this->rootUrl) {
```
ianbarber: I think we'll need to bump to 1.2 if we're being semver-ish, as this is potentially a BC break - in the past we would also use the configured basepath if one, but now we will override a configured base path with the one out of the service. Realistically I'm not sure this is really going to hit anyone though - WDYT? We should bump the version number in Client.php regardless to something.

tests/general/ResourceTest.php
```diff
@@ -0,0 +1,124 @@
...
+  {
+    parent::__construct($client);
+    $this->rootUrl = "";
+    $this->servicePath = "";
+    $this->version = "v1beta1";
+    $this->serviceName = "test";
+  }
+}
+
+class ResourceTest extends PHPUnit_Framework_TestCase
```
ianbarber: Nice test!

ianbarber:
For future people: anthony is a Googler, ignore clabot.

ianbarber:
Also: Test seems to be failing.

tests/general/ResourceTest.php
ircmaxell: Thanks!

src/Google/Service/Resource.php
ircmaxell: TBH, I could go either way. Considering that it's a completely undocumented parameter in public docs, and the only documented reference is in config.php:

      // Don't change these unless you're working against a special development
      // or testing environment.
      'base_path' => 'https://www.googleapis.com',

I could see it going either way. It's up to you.

I'll bump the Client.php version number in the PR as well.

PR #584, opened 2015-06-17
Add separate createBatch method to propagate batch URLs properly

ircmaxell:
Presently, batch URLs use the system configured global URL and a hard-coded batch path. This PR allows for services to specify the batch end point. 

It also fixes a problem with batches not correctly terminating headers when there is no body to the request.

### DIFF
--- a/src/Google/Http/Batch.php
+++ b/src/Google/Http/Batch.php
@@ -35,12 +35,15 @@ class Google_Http_Batch
 
   private $expected_classes = array();
 
-  private $base_path;
+  private $root_url;
 
-  public function __construct(Google_Client $client, $boundary = false)
+  private $batch_path;
+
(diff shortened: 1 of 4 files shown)


ircmaxell:
@erickoledadevrel The default configuration does not contain trailing slashes: https://github.com/google/google-api-php-client/blob/master/src/Google/Config.php#L52 therefore, I took the approach to strip the trailing slash [here](https://github.com/google/google-api-php-client/pull/584/files#diff-cd8ee972ef7dbf404f468f61a20852c0R45) and then add it at the combination step. I could add it back in the line above, but not really sure the benefit either way...

erickoledadevrel:
The `rootUrl` field in the discovery document does contains the slash by default, and since you are naming this field `root_url` instead of `base_path` it's probably better to have it respect the format of rootUrl and contain the trailing slash.

PR #586, opened 2015-06-25
Feature cs fixer

mjacobus:
Hello

I took the opportunity that the project has no open PR and ran php-cs-fixer on the project. I also added a script to automatically fix all the php code.

They adopted CS is the same as symfony's (Psr2 + enhancements) with the exception of a few their extra rules.

Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,6 @@ Run the PHPUnit tests with PHPUnit. You can configure an API key and token in Ba
 
     phpunit tests/
 
-Copy the ruleset.xml in style/ into a new directory named GAPI/ in your
-/usr/share/php/PHP/CodeSniffer/Standards (or appropriate equivalent directory),
-and run code sniffs with:
+Fix the code standard by running:
 
-        phpcs --standard=GAPI src/
+        ./bin/fix_code_standard.sh
--- a/autoload.php
+++ b/autoload.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * Copyright 2014 Google Inc.
  *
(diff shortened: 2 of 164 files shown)


mjacobus:
I signed it!

kesar:
It's hard to review a PR like this one ;-) 

Not sure if there is an issue and consensus about what CS rules is following this library.

ianbarber:
I very much like having the cs automatically fixable! The service classes are generated though, I would suggest excluding those and raising an issue for changes needed to the template. Given that the template changes are likely to be a while though, i would minimise the differences to the existing. The biggest one I can see is 4 space indent rather than 2 - could you alter to use a 2 space indent?

mjacobus:
@kesar yeah, I know. Besides this is a unsolicited PR =) You guys do not need to feel bad about not merging this ;). I do feel however, that it is important to follow the [CS](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md) widely adopted by the community, since they will be the consumers and potential contributors of the API.

@ianbarber  The goal was that [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) would fix indentation, but it apparently was not consistent. I can try to check if this is configurable to 2 spaces indentation, although I would strongly recommend using 4 to follow the community standards, as mentioned above.

If you guys are willing to merge I can try to make the changes this weekend and modify this PR or create a new one.

Cheers!

localheinz:
@mjacobus 

How about making this part of the build process? 

Here's an example from a different project:

* https://github.com/zendframework/modules.zendframework.com/blob/master/.travis.yml#L31

localheinz:
@mjacobus 

Never mind my example, we've already got a configuration for `phpcs` right here: https://github.com/google/google-api-php-client/blob/master/style/ruleset.xml

localheinz:
@ianbarber @kesar @mjacobus 

Do you want to take a look at #604?

bshaffer:
I would love to move the non-generated parts of the library onto PSR-1 and 2 code standards. However, we are merging `v2-dev` into master for the `2.0` release, and changing this now would be only make the merge process difficult and the diffs harder to read.

We will now track this task in #692

Issue #596, opened 2015-07-05
Multipart File Upload fails


bobertm:
I updated the google-api-php client lib. Since then the upload via insert fails with the following message.

Google_Service_Exception: Error calling POST https://www.googleapis.com//upload/drive/v2/files?uploadType=multipart: (400) Method Not Allowed in MyProjectPath\GoogleAPIClientLibPHP\src\Google\Http\REST.php on line 110

This is the code to upload a PDF File incl. meta data:

$client = new Google_Client();
$client->setAssertionCredentials($credentials);
if ($client->getAuth()->isAccessTokenExpired()) {
    $client->getAuth()->refreshTokenWithAssertion();
}
$service = new Google_Service_Drive($client);
$file = new Google_Service_Drive_DriveFile();
$file->setTitle("test.pdf");
$file->setMimeType('application/pdf');
$result = $service->files->insert(
        $file,
        array(
                'data' => file_get_contents("C:/Users/martin/Desktop/test.pdf"),
                'mimeType' => 'application/pdf',
                'uploadType' => 'multipart'
        )

With older Versions of the library it worked fine. The commit which seems to cause the problems is:
Drive.php, checked in on 22.06.2015 22:48:38

Has anyone an idea what is wrong or what has changed?

pdewouters:
I also experience the same issue.

ircmaxell:
Can you test the PR #597 to see if that rectifies the issue? (It should, and based on testing looks good). I just want to double-confirm prior to pushing.

bobertm:
bingo. #597 fixes the issue.

Thanks for your quick help.

ircmaxell:
I've merged that fix into master. Will check about pulling a release.

Thanks again!

pdewouters:
:+1: thanks!

tmatsuo:
Seems like we can close this

PR #597, opened 2015-07-06
Fix base component to trim trailing slash from the URL

ircmaxell:
Always trim the trailing slash off of the base component to prevent double slashes in the URL.


ircmaxell:
Verified as per #596, merging.

PR #599, opened 2015-07-07
Fix: Install dependencies in install section on Travis

localheinz:
This PR

* [x] moves installation of dependencies with Composer to the `install` section of the Travis configuration


localheinz:
I signed it, yo!

PR #602, opened 2015-07-07
Fix: Update Composer itself in before_install section

localheinz:
This PR

* [x] updates Composer itself in the `before_install` section on Travis


kesar:
:+1:

PR #603, opened 2015-07-07
Enhancement: Flatten directory structure, use PSR-4 autoloader

localheinz:
This PR

* [x] flattens the directory structure
* [x] switches to PSR-4 autoloading


bshaffer:
because of the larger changes for `2.0`, this task is now tracked in #692

PR #617, opened 2015-07-27
Stricter unix permissions for cache files.

jeanpaulgalea:
Check commit message for details.


jeanpaulgalea:
I signed it!

src/Google/Cache/File.php
```diff
@@ -186,6 +186,9 @@ private function acquireLock($type, $storageFile)
       );
       return false;
     }
+    if ($type == LOCK_EX) {
+      chmod($storageFile, 0600);
```
jeanpaulgalea: Preferably permission bits are set on `fopen`, but PHP doesn't allow that. I've looked into umask but that might introduce other issues since other scripts might be running under the same process.

bshaffer:
LGTM

PR #620, opened 2015-07-29
Fix an issue with custom headers, OAuth and Curl

andywooyay:
In Curl.php the headers are set first :

    curl_setopt($curl, CURLOPT_HTTPHEADER, $curlHeaders);

And then custom options set later:

    foreach ($this->options as $key => $var) {
      curl_setopt($curl, $key, $var);
    }

Unfortunately if `$this->options` contains any anything with $key = 10023 - even an empty array - it will override the headers set previously - e.g. `Authorization: bearer`. 

This little fix removes the headers from the options after setting them at the same time as the standard request headers.

### DIFF
--- a/src/Google/IO/Curl.php
+++ b/src/Google/IO/Curl.php
@@ -66,8 +66,10 @@ class Google_IO_Curl extends Google_IO_Abstract
       }
 
       if (isset($this->options[CURLOPT_HTTPHEADER])) {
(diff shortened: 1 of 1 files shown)


andywooyay:
I signed it!

bshaffer:
This seems easily avoidable by setting the headers on the `$request` object rather than in the `Google_Config` class.

Also, your PR should be rebased to a single commit.

Closing this for now.

Issue #608, opened 2015-07-09
Unsupported SSL context options are set. The following options are present, but have been ignored: cafile


nebadom:
Since earlier today (around 9:35AM EST), I'm getting this error message in the log:

```
  E 2015-07-09 10:35:51.967  200     779 B   889ms E 10:35:51.611 E 10:35:51.817 /get.php?u=xxxx@test.com&_=1436448941263
  190.188.222.26 - XXX [09/Jul/2015:06:35:51 -0700] "GET /get.php?u=xxxx@test.com&_=1436448941263 HTTP/1.1" 200 779 - "Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/44.0.2403.61 Safari/537.36" "mailroof-test.appspot.com" ms=889 cpu_ms=329 cpm_usd=0.000177 instance=00c61b117c0ffb0afc187ade6b0e941e8ff97f60 app_engine_release=1.9.24
  E 10:35:51.611 Unsupported SSL context options are set. The following options are present, but have been ignored: cafile
```

Still, the code seems to "work" in the sense that nothing is broken or missing in the functionality. But it is getting my log file full with this anoyment.

BTW, I was using PHP API 1.1.2 and today I upgraded to1.1.4 and later to the master version. I noticed the error log long after that and I thought that upgrade could be causing it. So I rolled back all the changes but still the error is there.

Any ideas of what could be causing this? and a solution?

PS: It seems to my like a ssl certificate error. Is it really possible?

eRadical:
Having the same issue:

[message:protected] => HTTP Error: Unable to connect: 'fopen(https://www.googleapis.com/bigquery/v2/projects/%PROJECT-ID%/jobs): failed to open stream: Unsupported SSL context options are set. The following options are present, but have been ignored: cafile
Fetch error'
    [string:Exception:private] => 
    [code:protected] => 2
    [file:protected] =>....................vendor/google/apiclient/src/Google/IO/Stream.php
    [line:protected] => 141

karensg:
I am not using any SSL certificates so I commented this line out of IO/Stream.php
`
$requestSslContext["cafile"] = dirname(__FILE__) . '/cacerts.pem';
`
That solved my issue for now until this problem is fixed.

ericuldall:
Also getting this issue.

jfradj:
Also getting this issue.
I don't think I made any changes that could lead to it.

[...]

Issue #570, opened 2015-05-27
Intermittent issue posting to Big Query from App Engine 


ericuldall:
We're getting this error occasionally when streaming rows into big query from GAE PHP:

Failed to post to BQ: HTTP Error: Unable to connect: 'fopen(**REDACTED_URL**): failed to open stream: Unsupported SSL context options are set. The following options are present, but have been ignored: cafile
Fetch error'

Steps to reproduce:
Sending 40 parallel requests of 40 records each containing 25 columns

dt-rush:
[Someone else with this issue seems to have found a googler who knows of the issue, has reproduced it, is working on it, and who provides a work-around](https://gae-php-tips.appspot.com/2013/12/23/getting-started-with-the-cloud-datastore-on-php-app-engine/).

At the very least, the error message looks the same. Can you test if the suggested work-around works for you?

adetogni:
Hi dt-rush. I have the same issue with SSL and I checked that link. However the solution proposed does not apply as at line 107 there is not that code and I couldn't find anywhere else in Stream.php.
It's very annoying as it happens randomly...

sjlangley:
This is an intermittent fetch error connecting to the backend - if possible add some retry logic to the write.

adetogni:
@sjlangley yep, that's an option but it's pretty time consuming as my app does a LOT of calls to Storage, in many files. I'm pretty sure I'll add some bugs when modifying such a big amount of code :)

bgold0:
This is happening to me too. Anyone who has found any other information please let me know

nebadom:
I have a open ticket on this subject:

https://github.com/google/google-api-php-client/issues/608

With no response so far.

bshaffer:
this was fixed in #631

PR #630, opened 2015-08-11
+ Make IO/Stream.php be App Engine aware

eRadical:
The issues were discussed in Case #608 and in Case #570


eRadical:
Sorry... googlebot was right I've used a wrong email address. Now it's fixed!

PR #631, opened 2015-08-11
+ Make IO/Stream.php be App Engine aware

eRadical:
The issues were discussed in Case #608 and in Case #570.

[sorry... the past pull request was messed up because I used a wrong address when I committed.]


4414:
http://apartmani-manjarema-fadic.hr/

bshaffer:
added in 7c191da

Issue #567, opened 2015-05-25
PubSub API not working.


s-chizhik:
Hi. I tried to use PubSub API with this client:

```php
<?php
require_once 'vendor/autoload.php';

// all authorizations things here

$client = new Google_Client();

$pub_sub = new Google_Service_Pubsub($client);

$topics = $pub_sub -> projects_topics -> listProjectsTopics('projects/my-project');

echo "Topics:" . PHP_EOL;
foreach ($topics as $topic) {
	echo "\t" . $topic['name'] . PHP_EOL;
}
```
When I tried to run it, I've got error
```
PHP Fatal error:  Uncaught exception 'Google_Service_Exception' with message 'Error calling GET https://www.googleapis.com/v1beta2/projects/my-project/topics: (404) Not Found' in /some_directory/vendor/google/apiclient/src/Google/Http/REST.php:110
```
By documentation I need to send requests to `pubsub.googleapis.com`, but here `www.googleapis.com`. I fixed this problem, but it's not solution:

```php
// Google_Service_Pubsub::_construct
    parent::__construct($client);
    $this->servicePath = 'v1beta2/';
    $this->version = 'v1beta2';
    $this->serviceName = 'pubsub';

    $this -> rootUrl = 'https://pubsub.googleapis.com';   // <= so ugly
    ...
```
Also, my composer.lock:
```json
 "name": "google/apiclient",
 "version": "1.1.4",
 "source": {
     "type": "git",
     "url": "https://github.com/google/google-api-php-client.git",
     "reference": "2adb5ba90612858d4add0342eee6b8b9aaca398d"
},
```

jonathantrevorh:
If you don't want to modify the source code to add the custom rootUrl, create an ini file with these contents:
```
base_path = 'https://pubsub.googleapis.com'
```

Then inject the configuration, by creating the client like so, instead:
```
$client = new Google_Client(new Google_Config($path_to_your_ini));
```

Once pubsub exits beta, it might be migrated to `www.google.com`, which would save us from this wokaround.

bshaffer:
We have merged in a fix for this, and `master` is now using `pubsub:v1`

Please update the library and confirm this fixes your problem. Thank you!

s-chizhik:
I've tested my code with `dev-master` revision. Seems it works fine! Would you be kind to tag this changes?

bshaffer:
I do not have the ability to tag this repo, unfortunately. @tbetbetbe are you the one to do this?

[...]

PR #637, opened 2015-08-17
updates pubsub from v1beta2 to v1

### LINKED ISSUE #618: Please update PubSub API from Beta1 to V1
Hi,

Pubsub seems to be out of beta. Can you update the code?

https://github.com/google/google-api-php-client/blob/master/src/Google/Service/Pubsub.php
$this->version = 'v1beta1';

Google API Reference V1:
https://cloud.google.com/pubsub/reference/rest/index#pubsub.googleapis.com_1

Please note that this api uses urls below https://pubsub.googleapis.com in v1, instead of https://www.googleapis.com/pubsub/v1beta1

bshaffer:
This is an attempt to fix #618 and #567 by upgrading the pubsub library to v1

### DIFF
--- a/src/Google/Service/Pubsub.php
+++ b/src/Google/Service/Pubsub.php
@@ -16,14 +16,14 @@
  */
 
 /**
- * Service definition for Pubsub (v1beta1).
+ * Service definition for Pubsub (v1).
  *
  * <p>
  * Provides reliable, many-to-many, asynchronous messaging between applications.</p>
  *
  * <p>
  * For more information about this service, see the API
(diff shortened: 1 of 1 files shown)


bshaffer:
I am a Google Employee, so I believe this should be good to go. Can someone confirm this?

RyanGordon:
It looks like some of the comments need to be updated with these changes: https://github.com/google/google-api-php-client/pull/637/files#diff-e5af3393c6631c01b5cff4ab345a7efbR551

Issue #639, opened 2015-08-18
CURLOPT SSLVERSION


gjellis:
Our environment
PHP 5.3.3
libcurl 7.19.7-46
google-api-php-client 1.1.5

Our problem
We had eight web heads all configured identical running the Google Client code.  After several weeks and escalated traffic, we suddenly started getting 'No Data Available' errors when we tried to execute a Google library.

After many hours of research, we realized we didn't have our code written properly.  Our order was off, calling OAuth Authorize before we created our Google Service Classes.  So, we fixed order and pushed to production.  Still getting 'No Data Available'.  After further research, we realized Curl was  causing httpd to die within src/Google/IO/Curl.php inside curl_exec().   

We finally had a break thru this morning on this issue.  We had tried many different debugging techniques, but everything we tried pointed back to the Apache child dying every time our application executed a curl_exec(), and all output stopped.
 
After a solid week of debugging and investigating, we have finally come up with a solution to this issue.
 
One character change in the Google API PHP Client code, inside the src/Google/IO/Curl.php,
 
we changed
    curl_setopt($curl, CURLOPT_SSLVERSION, 1);
to
    curl_setopt($curl, CURLOPT_SSLVERSION, 3);
 
And all was well.
 
There seemed to be a issue with this version of curl and https sites dealing with certificates.  We are not sure why one server never stopped working, and in the beginning, all the servers were working.  Something, might have been with the security certificates, that started causing the issue.
 
 
The StackOverFlow article that eventually lead us to the solution.
 
http://stackoverflow.com/questions/26009611/php-curl-returns-35-ssl-connect-error/32079279#32079279

eRadical:
This fix is rather worrisome...

What is actually means is that you are moving from CURL_SSLVERSION_TLSv1 (that is TLS v1.x) to CURL_SSLVERSION_SSLv3 which was deprecated recently (SSLv3 is disabled by default since version 7.39.0 or curl library).

At a given time in future (when PHP will catch up with the curl library) the code will probably error out.

**Have you tried commenting out the line?**... Thus leaving the curl library to decide what `ssl` version to use as this is the default behavior in curl (quote: CURL_SSLVERSION_DEFAULT » The default action. This will attempt to figure out the remote SSL protocol version.)

gjellis:
I did comment out this line and thing still worked.  So this must be the better fix.

Thank you

eRadical:
I'm sure in the long term will pay off.

Are you going to make a pull request out of it?
[...]

PR #644, opened 2015-08-26
Leave the unrelying curl library decide what SSL version to use

eRadical:
As discussed in case #639.

### DIFF
--- a/src/Google/IO/Curl.php
+++ b/src/Google/IO/Curl.php
@@ -73,8 +73,8 @@ class Google_IO_Curl extends Google_IO_Abstract
 
     curl_setopt($curl, CURLOPT_FOLLOWLOCATION, false);
     curl_setopt($curl, CURLOPT_SSL_VERIFYPEER, true);
-    // 1 is CURL_SSLVERSION_TLSv1, which is not always defined in PHP.
-    curl_setopt($curl, CURLOPT_SSLVERSION, 1);
+    // do not set this to deprecated SSL versions, best is to use underlying library default
+    //curl_setopt($curl, CURLOPT_SSLVERSION, CURL_SSLVERSION_DEFAULT);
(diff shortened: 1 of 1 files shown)


eRadical:
@bshaffer Any thoughts on this one?

bshaffer:
This code ensured that TLSv1 is used, which is a good thing because [SSLv3 is broken](https://googleonlinesecurity.blogspot.com/2014/10/this-poodle-bites-exploiting-ssl-30.html). 

So I don't think the justification is there for this change.

eRadical:
Worst case... there might be servers without TLS » as a radical response we can say "suits them well".

The problem is in negotiation as CURL_SSLVERSION_DEFAULT assures the negotiation from within the underlying curl library.

Setting a fixed protocol breaks negotiation as the connection will be tried only that version will be attempted.
PHP page recommends not setting any value for CURLOPT_SSLVERSION see: http://php.net/manual/en/function.curl-setopt.php

At http://curl.haxx.se/libcurl/c/CURLOPT_SSLVERSION.html it is mentioned that 
- SSLv2 is disabled by default since 7.18.1
- SSLv3 is disabled by default since 7.39.0
If someone has an older curl library then it means it is not interested that much in security anyway...

The other drawback is that setting it to 1 which is CURL_SSLVERSION_TLSv1 means that the library will only try with TLS 1.0... thus ignoring possible existing TLS 1.1 or TLS 1.2.

I still think it's a better option.

bshaffer:
These are good points. If you can rebase this PR to a single commit, I'll go ahead and merge it.

eRadical:
@bshaffer Is this ok?

bshaffer:
Whatever you did, you just added 3 more commits to this...

bshaffer:
fixed in a07f45

Issue #647, opened 2015-08-28
[exception] `Undefined variable: errorText` , code 0


s-chizhik:
In `Google_Auth_OAuth2::authenticate` in case when you got non json response, php will try to use undefined variable `$errorText` to make `Google_Auth_Exception`.

s-chizhik:
Anybody here?

PR #648, opened 2015-08-28
Fix Exception `Undefined variable: errorText` , code 0

s-chizhik:
Related to [this](https://github.com/google/google-api-php-client/issues/647)


s-chizhik:
@bshaffer, so what we're going to do? :grinning:

bshaffer:
Honestly, this is not a real issue that isn't worth fixing, as we have a major refactor coming in v2

PR #654, opened 2015-09-03
RFC #1: Refactor for adding UsableAuth library and GuzzleHttp

bshaffer:
## Major Changes
* Removes `Google_Http_Request` class
* Removes `Google_Http_CacheParser` class
* Removes `Google_IO_*` classes
* Removes `Google_Auth_AssertionCredentials` class
  * In favor of [`google-auth-library-php`](https://github.com/google/google-auth-library-php)
* Removes `Google_Auth_LoginTicket`
  * The payload of the JWT is sufficient
* JWT encoding methods
  * use [`firebase-jwt`](https://github.com/firebase/php-jwt) for this
* Removes `P12` support 
  * This is deprecated in favor of JSON credentials
* [`Google_Http_Batch`](https://github.com/bshaffer/google-api-php-client/blob/refactor-for-auth-and-guzzle-phase-1/src/Google/Http/Batch.php) is refactored to use Guzzle Batching

## Requires Consideration
* Tests - we need more of them!
* There are still some remaining Auth libraries we can remove / consolidate
  * `Simple`, `LoginTicket`, `AppIdentity`, and `ComputeEngine`
* Misc Caching
  * What about `Google_Cache_*`? 
* HTTP Caching
  * Do we need to build in support, or can the user just use the [Guzzle Cache Subscriber](https://github.com/guzzle/cache-subscriber)?
* Streaming 
  * Is Guzzle’s built-in streaming sufficient? (I vote yes)
* [`cacerts.pem`](https://github.com/google/google-api-php-client/blob/master/src/Google/IO/cacerts.pem) was removed
  * can we use [Guzzle's defaults](https://github.com/guzzle/RingPHP/blob/master/src/Client/ClientUtils.php#L23)? (may not be sufficient)
* GZIP
  * lots of logic surrounding GZIP - can we get rid of it?
  * more testing is needed for existing GZIP if not
* Retries
  * should we use the [Guzzle Retry Subscriber](https://github.com/guzzle/retry-subscriber)?
* Why are verifying for [more than one cert](https://www.googleapis.com/oauth2/v1/certs)?


bshaffer:
@jerjou

composer.json
```diff
@@ -6,7 +6,8 @@
     "homepage": "http://developers.google.com/api-client-library/php",
     "license": "Apache-2.0",
     "require": {
-        "php": ">=5.2.1"
+        "php": ">=5.4",
+        "google/auth": "dev-master"
```
bshaffer: once this branch is merged into `master`, we should tag the version of `google/auth` using [semantic versioning](http://semver.org/)

Issue #664, opened 2015-09-09
`Google_Service_Resource` lets `json_encode` fail silently


crishoj:
The `call` method in `Google_Service_Resource` relies on PHP's `json_encode` for the request body, which silently fails and returns `FALSE` in case of any non-UTF-8 encoded strings. 

Currently, errors are not handled. The `call` method proceeds as if nothing is wrong, leading to obscure problems later on, such as when `maybeMoveParametersToBody` in `Google_Http_Request` erroneously sets content type to `application/x-www-form-urlencoded` since there appears to be no request body.

crishoj:
I managed to duplicate the issue with pull request (#665). Apologies.

PR #665, opened 2015-09-09
Fail fast if JSON encoding fails

crishoj:
Let the `call` method fail fast in case `json_encode` fails to encode the request body, rather than proceeding with invalid data (#664).

### DIFF
--- a/src/Google/Service/Resource.php
+++ b/src/Google/Service/Resource.php
@@ -115,6 +115,8 @@ class Google_Service_Resource
             $this->convertToArrayAndStripNulls($parameters['postBody']);
       }
       $postBody = json_encode($parameters['postBody']);
+      if ($postBody === false && $parameters['postBody'] !== false)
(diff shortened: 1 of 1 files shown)


crishoj:
See #664 and https://groups.google.com/forum/#!topic/google-content-api-for-shopping/5I8MHpfbH4A for context.

bshaffer:
This pull request violates our [code standards for control structures](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#5-control-structures). You need to add curly braces so the change looks like this:

```php
if ($postBody === false && $parameters['postBody'] !== false) {
  throw new Google_Exception("JSON encoding failed. Ensure all strings in the request are UTF-8 encoded.");
}
```

crishoj:
Now code standards compliant.

bshaffer:
added in 635d6d8b7709a695aa69f5f659c51fc1776a8f13

PR #668, opened 2015-09-10
BUGFIX requiring of autoload.php in Google/Client.php

mojmirdurik:
Hi Tim,

working on youtube subscription in PHP using code [here] (https://developers.google.com/youtube/v3/code_samples/php#add_a_channel_subscription) I found a bug in Google PHP API library, in src/Google/Client.php:

```
if (!class_exists('Google_Client')) {
    require_once dirname(__FILE__) . '/autoload.php';
}

class Google_Client {...}
```

The `require_once dirname(__FILE__) . '/autoload.php';` is never executed as class Google_Client is defined in this file. I have just removed the condition.

BR

Mojo

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -15,9 +15,8 @@
  * limitations under the License.
  */
 
-if (!class_exists('Google_Client')) {
-  require_once dirname(__FILE__) . '/autoload.php';
-}
+require_once dirname(__FILE__) . '/autoload.php';
+
 
 /**
  * The Google API Client


mojmirdurik:
I signed it!)

2015-09-10 20:05 GMT+02:00 googlebot <notifications@github.com>:

bshaffer:
This is an innocuous change, but also it's not going to really fix anything either (If someone included `Client.php` manually then this change would prevent a fatal error, but we don't want to encourage that behavior anyway)

mojmirdurik:
ok :) Thank you for your feedback.

2015-10-09 19:45 GMT+02:00 Brent Shaffer <notifications@github.com>:

PR #672, opened 2015-09-14
Google_Http_MediaFileUpload::chunkSize make settable

dmitrydymarchuk:
I'm receiving chunked data with different size from some service and I would like to upload max available part of those chunks (multiple 262144 b),  but I can't set the `chunkSize` for every step of resumable upload.

### DIFF
--- a/src/Google/Http/MediaFileUpload.php
+++ b/src/Google/Http/MediaFileUpload.php
@@ -299,4 +299,9 @@ class Google_Http_MediaFileUpload
     $this->client->getLogger()->error($error);
     throw new Google_Exception($error);
   }
+
+  public function setChunkSize($chunkSize)
+  {
+    $this->chunkSize = $chunkSize;
+  }
 }


bshaffer:
@dmitrydymarchuk have you noticed you can specify chunk size [in the constructor](https://github.com/google/google-api-php-client/blob/master/src/Google/Http/MediaFileUpload.php#L81)?

dmitrydymarchuk:
Yes, I have, but I need to set different chunkSize for every step and recall constuctor for this is bad way.

bshaffer:
What is the use-case behind having to set a different chunkSize for each step?

max-voloshin:
@bshaffer it allows to use Google API more efficiently (without extra API calls) in case when we import different chunks of data from external sources and we don't know quantity and size of these chunks from the beginning.

bshaffer:
Have you tested and verified that setting the chunk size in this way effectively alters the HTTP request to allow for the functionality you're describing?

max-voloshin:
@bshaffer sure, we upload over 1 GB data to Google Analytics data set with different size of chunks.
Note: we don't have ability to buffer external data in order to use equal size of chunks.

Issue #658, opened 2015-09-04
Google_Http_MediaFileUpload::resumeUri needs to be public/gettable to resume from a different process


DavidAnderson684:
Explained in https://github.com/google/google-api-php-client/issues/143#issuecomment-40210269 - still a live issue for me.

I would create a patch with a getResumeUri() method, but there's already a method with that name that does something else. Ideas?

David

DavidAnderson684:
My code also has to set Google_Http_MediaFileUpload::progress in order to resume, so that has to be settable too.

bshaffer:
Fixed in #675

PR #673, opened 2015-09-15
Add methods to resume Uploads

tillz:
This Pull Request adds Support to resume an upload to google Drive, as asked in #658.
To achieve this, the following changes were Made:
* Renamed the `getResumeUri()`-Method to `fetchResumeUri()`, as it's more clear
* Added the Method `getResumeUri()` which is now used as a getter for the private `resumeUri`-property. 
* Added a `resume($uri)`-method to be called with an resumeUri, which will fetch the upload status & setup the object (set the resumeUri & progress properties).
* Moved code from `nextChunk()` to the new private `makePutRequest($request)` method to avoid duplicating code from `nextChunk()` in `resume()`.

With this changes, any chunked upload can be easily paused (or a better error-handling can be implemented) by simply getting the resumeURI from the `Google_Http_MediaFileUpload`-Object, storing it anywhere & later creating a new `Google_Http_MediaFileUpload`-Object with the same constructor-parameters as the original one & calling `$object->resume( $previously_stored_resumeUri);`, then proceeding with `nextChunk`. If `nextChunk` is called with chunked data, `$object->getProgress` can be used to find out the offset to be applied to the supplied data.

All informations are from [this reference](https://developers.google.com/drive/web/manage-uploads#resume-upload).

### DIFF
--- a/src/Google/Http/MediaFileUpload.php
+++ b/src/Google/Http/MediaFileUpload.php
@@ -126,35 +126,16 @@ class Google_Http_MediaFileUpload
   }
 
   /**
-   * Send the next part of the file to upload.
-   * @param [$chunk] the next set of bytes to send. If false will used $data passed
-   * at construct time.
-   */
-  public function nextChunk($chunk = false)
+  * Sends a PUT-Request to google drive and parses the response,
+  * setting the appropiate variables from the response()
+  *
+  * @param Google_Http_Request $httpRequest the Reuqest which will be send
+  *
+  * @return false|mixed false when the upload is unfinished or the decoded http response
+  *
+  */
+  private function makePutRequest(Google_Http_Request $httpRequest)
   {
-    if (false == $this->resumeUri) {
-      $this->resumeUri = $this->getResumeUri();
-    }
-
-    if (false == $chunk) {
-      $chunk = substr($this->data, $this->progress, $this->chunkSize);
-    }
-
-    $lastBytePos = $this->progress + strlen($chunk) - 1;
-    $headers = array(
-      'content-range' => "bytes $this->progress-$lastBytePos/$this->size",
-      'content-type' => $this->request->getRequestHeader('content-type'),
-      'content-length' => $this->chunkSize,
-      'expect' => '',
-    );
-
-    $httpRequest = new Google_Http_Request(
-        $this->resumeUri,
-        'PUT',
-        $headers,
-        $chunk
-    );
-
     if ($this->client->getClassConfig("Google_Http_Request", "enable_gzip_for_uploads")) {
       $httpRequest->enableGzip();
     } else {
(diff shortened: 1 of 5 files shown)


src/Google/Http/MediaFileUpload.php
```diff
@@ -263,7 +292,12 @@ public function getUploadType($meta)
     return self::UPLOAD_MULTIPART_TYPE;
   }
 
-  private function getResumeUri()
+  public function getResumeUri()
+  {
+    return ( $this->resumeUri !== null ? $this->resumeUri : $this->fetchResumeUri() );
```
bshaffer: This could better be written:

```php
return $this->resumeUri ?: $this->fetchResumeUri();
```

src/Google/Http/MediaFileUpload.php
```diff
@@ -165,7 +146,7 @@ public function nextChunk($chunk = false)
     $response->setExpectedClass($this->request->getExpectedClass());
     $code = $response->getResponseHttpCode();
     $this->httpResultCode = $code;
-
+
```
bshaffer: we need to get rid of unnecessary whitespace

src/Google/Http/MediaFileUpload.php
tillz: According to the README.md, PHP 5.2 is supported, but according to [the docs](http://php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary), this was introduced in 5.3?

bshaffer: You are right... we are still supporting 5.2 for this branch. Good catch.

src/Google/Http/MediaFileUpload.php
```diff
@@ -124,37 +124,18 @@ public function getHttpResultCode()
   {
     return $this->httpResultCode;
   }
-
+
```
bshaffer: Please trim this whitespace as well

bshaffer:
Thank you for this PR! Please trim the whitespace and rebase to a single commit, and we should be good to go here.

tillz:
Sorry, i'm not that familiar with rebasing. Can i just close this PR and send a new one, consisting of one commit?
Sidenote: the newlines on blank lines seem to be inconsistent in this repo. I think the reason of this is that the phpcs ruleset accepts whitespaces in blank lines. Possibly setting 'ignoreBlankLines' to false [here](https://github.com/google/google-api-php-client/blob/master/style/ruleset.xml#L52) would solve the problem (and save much work!)

bshaffer:
@tillz you can close this PR. You should never submit PRs from your master branch anyway, but instead should create a feature branch and work off that.

The following should accomplish this:
```
$ git remote add google git@github.com:google/google-api-php-client.git
$ git checkout -b patch-1 google/master
$ git cherry-pick f09b3d7
$ git push origin patch-1
```

Then you just need to submit a new PR from `tillz/patch-1` to `google/master`

PR #674, opened 2015-09-15
refactors tests to match class names

bshaffer:
* refactors tests to match class names
* removes `tests/general/AuthTest.php` because the functions being tested are part of the [firebase-jwt](https://github.com/firebase/php-jwt/) library, and not this library.
* removes `tests/testdata/*.pem`, `tests/testdata/*.p12`, and  `tests/testdata/*.json` because these were only being used by `tests/general/AuthTest.php`, which has been removed.


tbetbetbe:
LGTM, but before I merge, can you update the description to summarize why some files were deleted.

bshaffer:
@tbetbetbe good call. Updated. Would you like me to update the commit message as well?

tbetbetbe:
Yes, that would be nice as well.

bshaffer:
done

PR #675, opened 2015-09-15
Add methods to resume Uploads

tillz:
This Pull Request adds Support to resume an upload to google Drive, as asked in #658.
To achieve this, the following changes were Made:
* Renamed the `getResumeUri()`-Method to `fetchResumeUri()`, as it's more clear
* Added the Method `getResumeUri()` which is now used as a getter for the private `resumeUri`-property. 
* Added a `resume($uri)`-method to be called with an resumeUri, which will fetch the upload status & setup the object (set the resumeUri & progress properties).
* Moved code from `nextChunk()` to the new private `makePutRequest($request)` method to avoid duplicating code from `nextChunk()` in `resume()`.

With this changes, any chunked upload can be easily paused (or a better error-handling can be implemented) by simply getting the resumeURI from the `Google_Http_MediaFileUpload`-Object, storing it anywhere & later creating a new `Google_Http_MediaFileUpload`-Object with the same constructor-parameters as the original one & calling `$object->resume( $previously_stored_resumeUri);`, then proceeding with `nextChunk`. If `nextChunk` is called with chunked data, `$object->getProgress` can be used to find out the offset to be applied to the supplied data.

All informations are from [this reference](https://developers.google.com/drive/web/manage-uploads#resume-upload).

(Follow-Up to the closed PR #673)

### DIFF
--- a/src/Google/Http/MediaFileUpload.php
+++ b/src/Google/Http/MediaFileUpload.php
@@ -126,35 +126,16 @@ class Google_Http_MediaFileUpload
   }
 
   /**
-   * Send the next part of the file to upload.
-   * @param [$chunk] the next set of bytes to send. If false will used $data passed
-   * at construct time.
-   */
-  public function nextChunk($chunk = false)
+  * Sends a PUT-Request to google drive and parses the response,
+  * setting the appropiate variables from the response()
+  *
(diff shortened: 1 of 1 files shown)


bshaffer:
LGTM

PR #684, opened 2015-09-24
Library Update 1.1.5

bshaffer:
This change has been generated by a script that has detected changes in the

discovery doc of the API.

Check http://developers.google.com/discovery for more info.


bshaffer:
Something went wrong with the generator script [here](https://github.com/bshaffer/google-api-php-client/blob/bshaffer-update-test/src/Google/Service/Appengine.php#L1289), I will look into it

PR #685, opened 2015-09-25
Update "v2-dev" with "master"

bshaffer:


bshaffer:
as these have already been merged into `master`, we are good to go here on the authors.

PR #687, opened 2015-09-29
Added method to handle resumable uploads

maury91:
Example of how to use it :

//Check existence of a previous upload (if exists the last upload is interrupted)
if (file_exists('resume.url')) {
	$resumeUri = file_get_contents('resume.url');
	//Check if is resumable
	if ($media->resumeUpload($resumeUri)) 
		//Seek to the part to upload
		fseek($handle, $uploaded);
}

while (!$status && !feof($handle)) {
	$chunk = fread($handle, $chunkSizeBytes);
	$status = $media->nextChunk($chunk);
	//Save the resume url, it can change during upload
	file_put_contents('resume.url', $media->resumeUri);
}

//Delete resume url because the upload is complete
unlink('resume.url');

### DIFF
--- a/src/Google/Http/MediaFileUpload.php
+++ b/src/Google/Http/MediaFileUpload.php
@@ -45,7 +45,7 @@ class Google_Http_MediaFileUpload
   private $size;
 
   /** @var string $resumeUri */
-  private $resumeUri;
+  public $resumeUri;
 
   /** @var int $progress */
   private $progress;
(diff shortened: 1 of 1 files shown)


tillz:
Please have a look at #675, as it implements the same functionality.

maury91:
I signed it!

maury91:
tillz i've see your code, as you say they do the exact same thing, but i don't like the way how you have changed the name of a method "getResumeUri" -> "fetchResumeUri", if in someone else project exists a class that extend Google_Http_MediaFileUpload, this break his code.

I think that simply make public the variable resumeUri, is more clean

bshaffer:
@maury91 the method was previously private, so it's not possible for the change to break subclasses.

maury91:
Seems right, thanks @bshaffer for the info, if I've other patches to do I'll keep it mind

PR #688, opened 2015-10-02
Refactor auth phase 2

bshaffer:
This is the big one! Main points:

 - Removal of `Google_Auth_OAuth2`
   - Some functionality moved to `Google_AccessToken_Revoke` and `Google_AccessToken_Verify`
   - Some functionality moved to `Google_Client`
   - most functionality moved to `google/auth`
 - Removal of `Google_IO_*`
 - Logger changes
   - Removal of `Google_Logger_*`
   - Making all loggers accept `PSR\Log\Logger` interface
   - Using monolog for default
 - `Google_Client::authorize`
   - This now finds out what kind of `SubscriberInterface` you need based on the configuration of the client, and attaches the appropriate Guzzle HTTP Subscriber. Possible options:
      1. `Google\Auth\Simple` - for when `developer_key` is configured (i.e. `?key=abc`)
      1. `Google\Auth\ScopedAccessToken` - for when you've set the token using `$client->setAccessToken($token)`
      1. `Google\Auth\AuthTokenFetcher` - for when you have set `use_application_default_credentials` to true OR when your access token in step *b* needs refreshing.


bshaffer:
@tbetbetbe @surferjeffatgoogle @tmatsuo @silvolu

bshaffer:
A fix for the broken test above can be found [here](https://github.com/google/google-auth-library-php/pull/74)

composer.json
```diff
@@ -7,15 +7,20 @@
...
         "squizlabs/php_codesniffer": "~2.3"
     },
     "autoload": {
+        "psr-0": {
+            "Google_": "src/"
+        },
         "classmap": [
-            "src/"
+            "src/Google/Service/"
         ]
```
tmatsuo: It's rather a question. Why composer.json doesn't have GuzzleHttp as a dependency?

src/Google/AccessToken/Revoke.php
```diff
@@ -0,0 +1,73 @@
...
+    if (isset($token['refresh_token'])) {
+        $tokenString = $token['refresh_token'];
+    } else {
+        $tokenString = $token['access_token'];
+    }
+
+    $request = $this->http->createRequest('POST', Google_Client::OAUTH2_REVOKE_URI);
+    $request->addHeader('Cache-Control', 'no-store');
+    $request->addHeader('Content-Type', 'application/x-www-form-urlencoded');
+    $request->getBody()->replaceFields(array('token' => $tokenString));
```
tmatsuo: getBody()->replaceFields() doesn't seem straightforward.

Can you simply do as follows:

```php
$request = $this->http->createRequest('POST', Google_Client::OAUTH2_REVOKE_URI)
    ->addHeader('Cache-Control', 'no-store')
    ->addHeader('Content-Type', 'application/x-www-form-urlencoded')
    ->setPostField('token', $tokenString);
```
or something?

src/Google/AccessToken/Verify.php
```diff
@@ -0,0 +1,178 @@
...
+  {
+    // If we're retrieving a local file, just grab it.
+    if ('http' != substr($url, 0, 4)) {
+      $file = file_get_contents($url);
+      if ($file) {
+        return json_decode($file, true);
+      } else {
+        throw new Google_Auth_Exception(
+            "Failed to retrieve verification certificates: '" .
+            $url . "'."
```
tmatsuo: Maybe I'm wrong, but this string concatenation is little bit hard to read. Is it feasible to use String.format-ish thingy here?

tests/config/test.ini
```diff
@@ -1,7 +1,6 @@
...
 application_name = My Test application
 auth_class = Google_Auth_OAuth2
-[classes]
-Google_Auth_OAuth2[client_id] = 12345.apps.googleusercontent.com
-Google_Auth_OAuth2[client_secret] = gjfiwnGinpena3
-Google_Auth_OAuth2[redirect_uri] = http://example.com
\ No newline at end of file
+client_id = 12345.apps.googleusercontent.com
+client_secret = gjfiwnGinpena3
+redirect_uri = http://example.com
```
tmatsuo: Add a newline at the end?

tmatsuo:
Really a big one! I took a quick glance and put some comments.

composer.json
bshaffer: Because [`google/auth`](https://github.com/google/google-auth-library-php) has guzzle as a dependency. 

But you're right, we should add guzzle also. That way, if `google/auth` updates their guzzle version, it will cause a composer conflict rather than just breaking this library.

src/Google/AccessToken/Revoke.php
bshaffer: What we've done in other parts of the library is something like this:

```php
$request = $this->http->createRequest(
    'POST',
    Google_Client::OAUTH2_REVOKE_URI,
    [
        'body' => ['token' => $tokenString],
        'headers' => [
            'Cache-Control' => 'no-store',
            'Content-Type' => 'application/x-www-form-urlencoded',
        ],
    ]
);
```

This would accomplish the same as above.

src/Google/AccessToken/Verify.php
bshaffer: yeah I don't like it either. It should be:

```php
sprintf('Failed to retrieve verification certificates: "%s".', $url)
```

and also at the top it would be better as


```php
if (0 !== strpos($url, 'http')) {
   //...
}
```

composer.json
mtdowling: I noticed that library is using Guzzle version 5.x. 5.x is mostly in a maintenance mode at this point, and most development efforts are focused on version 6. Lots of libraries are adopting version 6 because it uses PSR-7 and has a much better story for async requests (for example, Drupal 8, Goutte, AWS SDK for PHP, etc.). I would strongly recommend that if you are building some new, that you use Guzzle 6 so that you're using something that will be maintained well into the future, and so that your libraries will be compatible with newer versions of other libraries in the ecosystem. Alternatively, you could support both versions, which helps to future proof your project. This is something we did in the AWS SDK for PHP: https://github.com/aws/aws-sdk-php/tree/master/src/Handler

bshaffer: Hey Matt! Thanks for the heads up. Yeah, I've noticed the newer docs are for 6. This was on my list but a lower priority. We will definitely look into the effort required to upgrade.

mtdowling: It's Michael :) and no problem. Feel free to reach out to me directly if you have any questions, or just mention me on any issues you want help with. I'm happy to help.

bshaffer: Sorry Michael! I saw "mt" and my brain filled in the blanks. 

We will make it a priority to switch to Guzzle 6 before tagging this release. Thanks for reaching out!

bshaffer: @mtdowling upgrading to v6 will take some consideration, as the entire underlying [google auth library](https://github.com/google/google-auth-library-php) is built on `GuzzleHttp\SubscriberInterface`.

I suppose I could convert the [subscriber classes](https://github.com/google/google-auth-library-php/blob/master/src/AuthTokenFetcher.php) to implement `_invoke()`, and then create the `HandlerStack` from that, but currently the idea is the `GuzzleHttp\Client` is passed in and "authenticated", but since `'handler'` is a constructor-only argument, this won't be possible.

The user must create the `GuzzleHttp\Client` and explicitly add our auth handlers, *or* we create a default client for them. So `$googleClient->getAuthHandler()` could return the callable for them to push on their stack. This is more work for the users than I would like to get a custom Http Client, but less magical than an `authorize` method. Is this the only way?

src/Google/Auth/AppIdentity.php
```diff
@@ -1,118 +0,0 @@
...
-use google\appengine\api\app_identity\AppIdentityService;
-
-if (!class_exists('Google_Client')) {
-  require_once dirname(__FILE__) . '/../autoload.php';
-}
-
-/**
- * Authentication via the Google App Engine App Identity service.
- */
-class Google_Auth_AppIdentity implements Google_Auth_Interface
```
bshaffer: My understanding is the functionality of this class will be added to [`google/auth`](https://github.com/google/google-auth-library-php)

src/Google/AccessToken/Revoke.php
tmatsuo: I still see the call for replaceFields...

4414:
Welcome in Komiža, small touristic place on island Vis in place Komiža.
http://apartmani-manjarema-fadic.hr/

PR #690, opened 2015-10-05
composer: bump to PHPUnit ~4.0

TomasVotruba:


TomasVotruba:
I signed it!

TomasVotruba:
Sure :)

bshaffer:
What is the justification for this change? Can you explain why you'd like us to bump the PHPUnit version?

TomasVotruba:
Thanks for the question. I guess it's the same reason as for using Windows 8 over Windows 95.

PHPUnit 5.0 is now out and using deprecated packages makes it more complicated to develop that otherwise.

bshaffer:
I agree that we would be better off on PHPUnit 4 (PHPUnit 5 requires php 5.5, so we can't hop on that quite yet).

We will integrate this change into [v2-dev](https://github.com/google/google-api-php-client/tree/v2-dev) since that will be merged into master soon. Thanks for this!

TomasVotruba:
Ok, thanks.

Issue #696, opened 2015-10-14
id_token verification fails unexpectedly with Google Sign-In SDK for iOS


ecoopnet:
The id token id_token that is created by Google iOS SDK( https://cocoapods.org/pods/GoogleSignIn ), always fails.

This is because the issuer of the id_token is "https://accounts.google.com', and

> Google_Auth_OAuth2#verifyIdToken($idToken)

checks only "accounts.google.com" by default.

In All Google Sigin-Inl Guides(web/android/iOS), 
the valid issuer identifier are accounts.google.com and https://accounts.google.com for Google ID tokens like below, 
so I think verifyIdToken() should accept both "https://accounts.google.com' and "accounts.google.com".

```
Verify that the value of iss in the ID token is equal to accounts.google.com or https://accounts.google.com.
```

```
iss always  The Issuer Identifier for the Issuer of the response. Always accounts.google.com or https://accounts.google.com for Google ID tokens.
```
- Google Sigin-In SDK for iOS (uses accounts.google.com )
  https://developers.google.com/identity/sign-in/ios/backend-auth
- Google Sign-In for Android(uses https://accounts.google.com ).
  https://developers.google.com/identity/sign-in/android/backend-auth
- Google Sign-In for Websites
  https://developers.google.com/identity/sign-in/web/backend-auth

PR #697, opened 2015-10-14
Accept documented issuers accounts.google.com and https://accounts.google.com

ecoopnet:
changed to accept both issuers "accounts.google.com" and "https://accounts.google.com/"
to fix the issue #696.

https://developers.google.com/identity/sign-in/web/backend-auth
- Verify the integrity of the ID token
```
The value of iss in the ID token is equal to accounts.google.com or https://accounts.google.com.
```

### DIFF
--- a/src/Google/Auth/OAuth2.php
+++ b/src/Google/Auth/OAuth2.php
@@ -32,6 +32,7 @@ class Google_Auth_OAuth2 extends Google_Auth_Abstract
   const AUTH_TOKEN_LIFETIME_SECS = 300; // five minutes in seconds
   const MAX_TOKEN_LIFETIME_SECS = 86400; // one day in seconds
   const OAUTH2_ISSUER = 'accounts.google.com';
+  const HTTPS_OAUTH2_ISSUER = 'https://accounts.google.com';
 
   /** @var Google_Auth_AssertionCredentials $assertionCredentials */
   private $assertionCredentials;
@@ -488,7 +489,7 @@ class Google_Auth_OAuth2 extends Google_Auth_Abstract
       $audience = $this->client->getClassConfig($this, 'client_id');
     }
 
-    return $this->verifySignedJwtWithCerts($id_token, $certs, $audience, self::OAUTH2_ISSUER);
+    return $this->verifySignedJwtWithCerts($id_token, $certs, $audience, array(self::OAUTH2_ISSUER, self::HTTPS_OAUTH2_ISSUER));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ecoopnet:
I signed it!
My company name is "tech vein, Inc.".

ecoopnet:
I signed it!
My company name is "tech vein, Inc.".

src/Google/Auth/OAuth2.php
```diff
@@ -509,6 +515,11 @@ public function verifySignedJwtWithCerts(
       $issuer = null,
       $max_expiry = null
   ) {
+    // accept multiple issuers with array.
+    $issuers = null;
+    if ($issuer) {
+      $issuers = is_array($issuer) ? $issuer : array($issuer);
+    }
```
bshaffer: This chunk of code can be removed, and the next line below should be:

```php
if ($issuer && !in_array($iss, (array) $issuer)) {
```

bshaffer:
Thanks for tracking this down, submitting the PR, and providing all the documentation! It looks good.

I've added some fixes. We also need to add tests. I can do both these things, but it's probably best if you do it since it's your PR.

src/Google/Auth/OAuth2.php
```diff
@@ -32,6 +32,7 @@ class Google_Auth_OAuth2 extends Google_Auth_Abstract
   const AUTH_TOKEN_LIFETIME_SECS = 300; // five minutes in seconds
   const MAX_TOKEN_LIFETIME_SECS = 86400; // one day in seconds
   const OAUTH2_ISSUER = 'accounts.google.com';
+  const HTTPS_OAUTH2_ISSUER = 'https://accounts.google.com';
```
bshaffer: change to `OAUTH2_ISSUER_HTTPS`

src/Google/Auth/OAuth2.php
```diff
@@ -596,12 +607,12 @@ public function verifySignedJwtWithCerts(
     }
 
     $iss = $payload['iss'];
-    if ($issuer && $iss != $issuer) {
+    if ($issuers && !in_array($iss, $issuers)) {
```
bshaffer: I would like to see a comment here

```php
// support HTTP and HTTPS issuers
// @see https://developers.google.com/identity/sign-in/web/backend-auth
```

ecoopnet:
Thanks @bshaffer. I added your fix to this PR.
Please add tests.

bshaffer:
added in 043fab574f4921e8693b60bb49d74e65de8e6b6b

PR #715, opened 2015-10-28
uses guzzle StreamHandler by default [AppEngine only]

bshaffer:
This is required for AppEngine calls to a succeed, otherwise the connection error `Curl Errno 7` is thrown.

We could only do this for when `isAppEngine` is true, but I'm a fan of as few conditional checks as possible. Are there any downsides to defaulting to this StreamHandler over Curl?

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -27,9 +27,10 @@ use Google\Auth\UserRefreshCredentials;
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Collection;
+use GuzzleHttp\Ring\Client\StreamHandler;
 use Psr\Log\LoggerInterface;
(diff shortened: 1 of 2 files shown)


bshaffer:
advantages: does not require `curl.so`, runs on appengine
disadvantages: ??

tmatsuo:
Seems like App Engine now supports cURL:
https://cloud.google.com/appengine/docs/php/#PHP_cURL_support

Have you tried that?

tmatsuo:
Sorry, apparently you tried them. I think we should make it (App Engine cURL support) work also.

bshaffer:
@tmatsuo yes, AppEngine supports cURL, but I was unable to get this library to work even so. I agree it would be better to fix the issue, but this is a good solution in the interim.

PR #726, opened 2015-11-02
Update MediaFileUpload.php

tibee:
Typo in source.


tibee:
I signed it!

bshaffer:
When using double quotes for string values, array keys are referenced without the quotes.

tibee:
Certainly not.

It makes PHP Notice.
See manual:
http://en.php.net/manual/en/language.types.array.php#language.types.array.foo-bar

```
PHP Notice:  Use of undefined constant domain - assumed 'domain'.
PHP Notice:  Use of undefined constant message - assumed 'message'.
```

tibee:
Correct link:
http://www.php.net/manual/en/language.types.array.php#language.types.array.foo-bar

thejaswip:
i agree with tibee ! please kindly review

PR #728, opened 2015-11-03
Don't try to convert to model if alt=media present

sqrrrl:
One more fix for media downloads -- don't set the expected return class so that the raw body is returned.

### DIFF
--- a/src/Google/Service/Resource.php
+++ b/src/Google/Service/Resource.php
@@ -130,7 +130,7 @@ class Google_Service_Resource
         $this->stackParameters,
         $method['parameters']
     );
-    
+
     foreach ($parameters as $key => $val) {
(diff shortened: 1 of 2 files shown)


bshaffer:
Thank you for this! Question - shouldn't these types of API calls not even set an expected class? Is there somewhere in our documentation we can point for this use-case?

sqrrrl:
The alt parameter is, literally, to request an alternate representation. The default is JSON, but some APIs like Drive and others have resources that have multiple representations -- JSON for metadata, blobs for content. Same for storage and a few others.

In those cases, the methods do have an expected class when the default JSON representation is used. But for media representations, it doesn't make sense to use them.

SurferJeffAtGoogle:
Can we update a test to exercise this new code path?

bshaffer:
merged in #732

PR #729, opened 2015-11-03
adds composer support in README and adds autoload back

bshaffer:
This is to address [user concerns](http://stackoverflow.com/questions/33483326/how-do-i-use-composer-to-setup-google-api-php-client) regarding composer support. 

In addition, we will be adding a downloadable zip file to each project that includes all the dependencies, for those users who would prefer to not use composer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,38 @@ http://developers.google.com/api-client-library/php
 
 ## Installation ##
 
-For the latest installation and setup instructions, see [the documentation](https://developers.google.com/api-client-library/php/start/installation).
+You can use **Composer** or simply **Download the Release**
+
+### Composer
+
+The preferred method is via [composer](https://getcomposer.org). Follow the
+[installation instructions](https://getcomposer.org/doc/00-intro.md) if you do not already have
+composer installed.
+
+Once composer is installed, execute the following command in your project root to install this library:
+
+```sh
+composer require google/apiclient:^2.0.0@RC
+```
+
+Finally, be sure to include the autoloader:
+
+```php
+require_once '/path/to/your-project/vendor/autoload.php';
+```
+
+### Download the Release
+
+If you abhor using composer, you can download the package in its entirety. The [Releases](https://github.com/google/google-api-php-client/releases) page lists all stable versions. Download any file
+with the name `google-api-php-client-[RELEASE_NAME].zip` for a package including this library and its dependencies.
+
+Uncompress the zip file you download, and include the autoloader in your project:
+
+```php
+require_once '/path/to/google-api-php-client/vendor/autoload.php';
+```
+
+For additional installation and setup instructions, see [the documentation](https://developers.google.com/api-client-library/php/start/installation).
 
 ## Basic Example ##
 See the examples/ directory for examples of the key client features. You can
--- a/src/Google/autoload.php
+++ b/src/Google/autoload.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * THIS FILE IS FOR BACKWARDS COMPATIBLITY ONLY
+ *
+ * If you were not already including this file in your project, please ignore it
+ */
+
+$file = __DIR__ . '/../../vendor/autoload.php';
+
+if (!file_exists($file)) {
+  $exception = 'This library must be installed via composer or by downloading the full package.';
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -16,7 +16,34 @@ http://developers.google.com/api-client-library/php
...
+
+### Composer
+
+Install this library via [composer](https://getcomposer.org) by running the following command:
+
+```sh
+# change directories to the root of your project
+cd /path/to/my-project
+# if you don't already have composer, install it
+curl -sS https://getcomposer.org/installer | php
```
tmatsuo: I would defer it to the guide on the composer website.

README.md
```diff
@@ -16,7 +16,34 @@ http://developers.google.com/api-client-library/php
...
+# install this library
+./composer.phar require google/apiclient
```
tmatsuo: There are multiple way to install composer and the command line varies. How about just 

```sh
$ composer require google/apiclient
```
?

README.md
```diff
@@ -16,7 +16,34 @@ http://developers.google.com/api-client-library/php
...
+
+> The above example adds a file `composer.phar` to your project working directory for simplicity. We recommend
+> you [install composer globally](https://getcomposer.org/doc/00-intro.md#globally) for local development.
+
+### Autoloading
+
+Once installed, be sure to include the autoloader:
+
+```php
+require_once '/path/to/google-api-php-client/vendor/autoload.php';
```
tmatsuo: Is the path `/path/to/google-api-php-client` correct?

tmatsuo: I would say `/path/to/your-project`

README.md
tmatsuo: Here is the reaosn why.

Well, curling a script from the web and pipe it into an interpreter, might be a trend, but in my opinion, it's dangerous in some senses. When composer website started to suggest more secure way, this document will provide an orphaned, insecure instruction.

bshaffer: I am really trying to keep this guide simple. As the stack overflow thread shows, some users are entirely unfamiliar with composer. Sending them to the website for installation will just increase the bounce rate. That is why I went this way... to get them up and going the fastest.

README.md
bshaffer: that only works if they've taken the time to install composer globally. If they've done that, I assume they know the difference (and I mention it in the quote below)

README.md
bshaffer: yeah, hmm... I mean't it specifically for those who downloaded the release w/o composer. But you're right, it's confusing.

README.md
tmatsuo: Well, people who don't know composer can still download the zip file, right?

bshaffer: Yes, but composer should still be encouraged, hence the example.

tmatsuo: On https://getcomposer.org/doc/00-intro.md
There is a mention about Windows installer, which is better than the command line we are showing. Does the command line work with Windows console?

bshaffer: no clue, but I think you're right... pointing to the composer website is a better long term solution, and will make for a cleaner readme

bshaffer:
@tmatsuo PTAL

src/Google/autoload.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$file = __DIR__ . '/../../vendor/autoload.php';
```
kesar: vendor could have a different folder name ( https://getcomposer.org/doc/06-config.md#vendor-dir )

tmatsuo: Good point :)
Can you use __DIR__ . '/../autoload.php' instead?

bshaffer: @kesar this is a deprecated file meant to move users of v1.x over to 2 with minimal difficulty. If a user has a custom `vendor` directory, they should not be using this file anyway, so not think it's worth handling this edge case.

@tmatsuo I am unclear what that would accomplish.

tmatsuo: > I am unsure what that would accomplish.

@bshaffer I think it should work even you use custom `vendor` directory. I may be wrong though.

tmatsuo: Ah, OK. I misunderstand that `src/Google/autoload.php` is installed when you're using composer.

tmatsuo: Alright. LGTM

PR #732, opened 2015-11-04
Additional Support for alt=media

bshaffer:
Adds tests and exception handling for #728

### DIFF
--- a/tests/Google/Service/ResourceTest.php
+++ b/tests/Google/Service/ResourceTest.php
@@ -18,6 +18,10 @@
  * under the License.
  */
 
+use GuzzleHttp\Message\Request;
+use GuzzleHttp\Message\Response;
+use GuzzleHttp\Stream\Stream;
+
 class Test_Google_Service extends Google_Service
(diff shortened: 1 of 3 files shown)


bshaffer:
@sqrrrl @tmatsuo PTAL

sqrrrl:
FYI - Already updated #728 with some additional tests, but they're fairly superficial.

tests/Google/Service/ResourceTest.php
```diff
@@ -193,4 +197,93 @@ public function testAppEngineSslCerts()
...
+          "testMethod" => array(
+            "parameters" => array(),
+            "path" => "method/path",
+            "httpMethod" => "POST",
+          )
+        )
+      )
+    );
+
+    $decoded = $resource->call('testMethod', $arguments);
```
sqrrrl: Should pass in an expected class here. One of the key things is ignoring the expected class when alt=media is present. Not specifying one in the first place can mask regressions.

PR #735, opened 2015-11-05
ensures json errors are not thrown

bshaffer:


src/Google/Http/REST.php
```diff
@@ -93,10 +94,21 @@ public static function decodeHttpResponse(
...
 
     // return raw response when "alt" is "media"
     $isJson = !($request && 'media' == $request->getQuery()->get('alt'));
-    $result = $isJson ? $response->json() : $body;
+
+    // set the result to the body if it's not set to anything else
+    if ($isJson) {
+      try {
+        $result = $response->json();
+      } catch (ParseException $e) {
```
tmatsuo: Now the exception is just thrown away. Is it possible to log details as a warning?

bshaffer: The assumption is the exception indicates the response *isn't* JSON, rather than indicating an actual error.

bshaffer: Also, whatever the non-JSON is gets returned, so nothing super valuable is thrown away (IMHO), just whatever parse error was discovered by `json_decode`.

tmatsuo: I see. Thanks.

tmatsuo:
LGTM

PR #737, opened 2015-11-06
fixes batch requests for POST

### LINKED ISSUE #736: Request body empty when using batch - V2 beta API
Hi,

When batching Gmail API requests, such as "modify" requests for messages (setting/unsetting labels on messages), I noticed that with v2 beta API, the API returns a 500 server error.

By using some print_r statements in the v2 API source code, I have found that when using batch (Google_Http_Batch) containing POST requests with a body, the request effectively sent does not have any body. However, when NOT using batch calls, everything works fine. Further debugging shows that the request being batched is well constructed, and contains a body.

Having a look at the Google\Http\Batch.php source code (lines 77-81) :       
$body .= sprintf(
          '%s %s HTTP/%s',
          $request->getMethod(),
          $request->getResource(),
          $request->getProtocolVersion()
);

we can see that there's nothing to include the $request body (could be obtained by $request->getBody()).

It tried to figure out how it could be done (by refering to the request being sent by v1 API), with :
      $body .= sprintf(
          '%s %s HTTP/%s %s',
          $request->getMethod(),
          $request->getResource(),
          $request->getProtocolVersion(),
          $request->getBody() ? "\n\n" . $request->getBody():''
);

but I always encounter a 500 server error.

Please, anyone can have a look at this issue ? Thanks

bshaffer:
fixes #736

### DIFF
--- a/src/Google/Http/Batch.php
+++ b/src/Google/Http/Batch.php
@@ -61,26 +61,38 @@ class Google_Http_Batch
   {
     $body = '';
     $classes = array();
+    $batchHttpTemplate = <<<EOF
+--%s
+Content-Type: application/http
+Content-Transfer-Encoding: binary
+MIME-Version: 1.0
+Content-ID: %s
+
+%s%s
+%s
+
+
+EOF;
 
     /** @var Google_Http_Request $req */
     foreach ($this->requests as $key => $request) {
-      $request->addHeaders(
-          [
-            'Content-Type' => 'application/http',
-            'Content-Transfer-Encoding' => 'binary',
-            'MIME-Version' => '1.0',
-            'Content-ID' => $key,
-          ]
-      );
-      $body .= "--{$this->boundary}";
-      $body .= Request::getHeadersAsString($request) . "\n\n";
-      $body .= sprintf(
+      $firstLine = sprintf(
           '%s %s HTTP/%s',
           $request->getMethod(),
           $request->getResource(),
           $request->getProtocolVersion()
       );
-      $body .= "\n\n";
+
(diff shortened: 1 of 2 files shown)


src/Google/Http/Batch.php
```diff
@@ -61,26 +61,38 @@ public function execute()
...
-      $body .= "\n\n";
+
+      $content = (string) $request->getBody();
+
+      $body .= sprintf(
+          $batchHttpTemplate,
+          $this->boundary,
+          $key,
+          $firstLine,
+          Request::getHeadersAsString($request),
```
tmatsuo: Does `Request::getHeadersAsString()` have a newline at the beginning? Otherwise it will be in the same line with the `$firstLine` as far as I read the code.

bshaffer: Yes it does... it's very counter intuitive, and I don't like it, but it does.

We could also call it with `trim` and then the template could have a new line. I'm not a particular fan of either way.

tmatsuo: The implementation is:
https://github.com/guzzle/guzzle/blob/5.3/src/Message/AbstractMessage.php#L206

Is it safe to assume they won't change the implementation?

bshaffer: yeah I think so :D

tmatsuo:
LGTM

src/Google/Http/Batch.php
bshaffer: I understand why they are doing this - if there are no headers, then the first line should still only have one new-line between it and the body, so this prevents an unnecessary newline character if there are no headers.

PR #751, opened 2015-11-16
Update UPGRADING.md

ruslan-polutsygan:
Seems header needs to be distinguished


ruslan-polutsygan:
I signed it!

bshaffer:
This is as expected. The section for impersonating a user is supposed to be under the removal of `AssertionCredentials`.

Are you proposing this fix because you found the current way confusing?

PR #762, opened 2015-11-30
Add strategy constants for PageSpeed

bocharsky-bw:


bshaffer:
this file is auto-generated, so this change would be overwritten when we update next.

PR #763, opened 2015-11-30
Make property protected and add getter for it

bocharsky-bw:
Make `pagespeedapi` property protected and add public getter for it

### DIFF
--- a/src/Google/Service/Pagespeedonline.php
+++ b/src/Google/Service/Pagespeedonline.php
@@ -33,7 +33,7 @@ class Google_Service_Pagespeedonline extends Google_Service
 {
 
 
-  public $pagespeedapi;
+  protected $pagespeedapi;
   
 
   /**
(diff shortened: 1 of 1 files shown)


Nyholm:
:+1:

bshaffer:
These files are auto generated so this change would just be overwritten on the next update

bocharsky-bw:
@bshaffer So how to do this changes and avoid overwriting on the next update?

bocharsky-bw:
Well, for now I just want to improve code quality here.

bshaffer:
I see. I agree with the change, but as this is auto-generated, the work for this change may not be worth the investment. However, the code to generate these classes can be found [here](https://github.com/google/apis-client-generator) if you're interested in implementing a more permanent fix.

bocharsky-bw:
Yea, it'll be a bit hard this way. Thanks for explanation how it works.

PR #765, opened 2015-12-03
Remove a useless isset

OlivierBarbier:
No need to call isset twice.


OlivierBarbier:
I signed it!

quantumpacket:
```php
if (isset($result['error'], $result['error']['errors'])) {
```
Should just be:
```php
if (isset($result['error']['errors'])) {
```

For `$result['error']['errors']` to exist it already implies that `$result['error']` would already exist. Right?

OlivierBarbier:
Right.

bshaffer:
agreed, let's go with @quantumpacket's solution.

bshaffer:
closing, this will be refactored out soon

Issue #769, opened 2015-12-09
Parameter in unlock method not used


jordikroon:
Not sure if it's their for a reason.

https://github.com/google/google-api-php-client/blob/master/src/Google/Cache/File.php#L215-L220

PR #771, opened 2015-12-11
Remove unused parameter $storageFile in unlock method

jordi12100:


jordi12100:
I signed it!

jordi12100:
Fixes #769

bshaffer:
We need to remove the parameter from [calls that are still passing it in](https://github.com/google/google-api-php-client/blob/master/src/Google/Cache/File.php#L84).

jordi12100:
Yes you're right. I've removed the parameters from the usages.

bshaffer:
LGTM!

caleblloyd:
RE: the change made on: https://github.com/google/google-api-php-client/blob/master/src/Google/Cache/File.php#L76

it seems to me that this statement now has no effect, since the output is not assigned to a variable. did you mean to leave it as:

    $data =  unserialize($data);

bshaffer:
thank you @caleblloyd! This is now fixed in 20e6a625a0dbf0f41815baa565549d9af39c5586

PR #790, opened 2015-12-30
improves Model construction by allowing nested properties through arrays

bshaffer:
Makes the following array-passing in the constructor and manual class creation identical:

*\* BEFORE **

``` php
$query = new \Google_Service_Datastore_Query();

$order = new \Google_Service_Datastore_PropertyOrder(['direction' => 'descending']);
$property = new \Google_Service_Datastore_PropertyReference(['name' => 'created']);
$order->setProperty($property);
$kind = new \Google_Service_Datastore_KindExpression(['name' => self::KIND]);

$query->setOrder([$order]);
$query->setKinds([$kind]);
$query->setLimit($limit);
```

*\* AFTER **

``` php
$query = new \Google_Service_Datastore_Query([
    'order' => [
        [
            'direction' => 'descending',
            'property'  => [ 'name' => 'created' ],
        ],
    ],
    'kinds' => [
        [ 'name' => self::KIND ],
    ],
    'limit' => $limit,
]);
```

### DIFF
--- a/src/Google/Collection.php
+++ b/src/Google/Collection.php
@@ -15,31 +15,31 @@ class Google_Collection extends Google_Model implements Iterator, Countable
 
   public function rewind()
   {
-    if (isset($this->modelData[$this->collection_key])
-        && is_array($this->modelData[$this->collection_key])) {
-      reset($this->modelData[$this->collection_key]);
+    if (isset($this->{$this->collection_key})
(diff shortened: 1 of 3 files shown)


bshaffer:
As this is a big change for very small noticeable gain, this will be merged in after the tagged stable release of `v2.0.0`

PR #794, opened 2016-01-04
GAE: Removed usage of /etc/ca-certificates.crt

zyberspace:
Removed a `StreamHandler` config line that caused my appengine to log a notice every time i made an api call. ("Unsupported SSL context options are set. The following options are present, but have been ignored: allow_self_signed, cafile")
The reason for this is that guzzle [sets](https://github.com/guzzle/guzzle/blob/master/src/Handler/StreamHandler.php#L347-L371) the `cafile` and `allow_self_signed` options when you use the `verify`-option and both are not supported by GAE.

Removed the config line and everything works fine now. Maybe it was needed in a past version of the GAE?

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -1011,7 +1011,6 @@ class Google_Client
       if ($this->isAppEngine()) {
         // set StreamHandler on AppEngine by default
         $options['handler']  = new StreamHandler();
-        $options['defaults']['verify'] = '/etc/ca-certificates.crt';
       }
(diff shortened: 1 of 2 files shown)


zyberspace:
Signed it.

zyberspace:
Ok, seems like this didn't fix anything at all. Don't know why but these notices are back even it worked like an hour ago.
Will close this pull request now and maybe open a second one if i will ever fix this issue.

PR #802, opened 2016-01-14
GAE: Workaround for unsupported ssl options

zyberspace:
Guzzle sets some default ssl context options that are not supported by the App Engine and are creating an error in the log every time an api gets called.
The app will still work but it makes filtering the logs for "real" errors pretty hard.

As said in the [App Engine docs](https://cloud.google.com/appengine/docs/php/urlfetch/#PHP_Secure_connections_and_HTTPS) the App Engine proxy takes care about certificate verification so we can safely set all the ssl context option to `null`.

The errors i get **without** this workaround:

    Unsupported SSL context options are set. The following options are present, but have been ignored: cafile
    Unsupported SSL context options are set. The following options are present, but have been ignored: cafile
    Unsupported SSL context options are set. The following options are present, but have been ignored: cafile

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -1011,7 +1011,12 @@ class Google_Client
       if ($this->isAppEngine()) {
         // set StreamHandler on AppEngine by default
         $options['handler']  = new StreamHandler();
-        $options['defaults']['verify'] = '/etc/ca-certificates.crt';
+        // we overwrite this but guzzle will do a lot of unneeded work if this is not set to false
+        $options['defaults']['verify'] = false;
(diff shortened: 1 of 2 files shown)


zyberspace:
Ok, i really messed up here, sry. So the original line only served guzzle 5 while i was using guzzle 6.
So these changes did actually nothing in guzzle 6 even it worked in my tests.

Also Guzzle 5 throws an exception when this gets used in production (even the tests are working, because i relied on them to test this in guzzle 5 before i installed guzzle 5 in my production app).

So i am gonna close this now and search for a way to remove these ssl context options in guzzle 6 even it doesn't seem like this is possible without some big changes.

PR #808, opened 2016-01-16
Handle exceptions correctly when using Guzzle 5

colinfrei:
When calls cause an exception the error response is passed to `decodeHttpResponse` method directly. When using Guzzle 5, that response is not a PSR-7 response, which causes an error.

This pull requests catches that and converts it if necessary.


thejaswip:
+1

funivan:
+1

bshaffer:
@colinfrei This is great! Thanks for catching and fixing this.

Issue #809, opened 2016-01-19
Reduce/Eliminate manual steps in OAuth flow


vsubramani:
When using the php libraries for obtaining OAuth credentials to use with APIs hosted on GCP, these manual steps are involved:
1. Copy Auth url to browser
2. Accept on the consent screen
3. Copy the auth code into the go application

See example https://developers.google.com/google-apps/calendar/quickstart/php

Google API client libraries for other languages host an http listener thereby avoiding the need for  manual steps 1 and 3. Please consider doing similar for the php client libraries.

bshaffer:
This may be more complicated than for python, due to the lack of multithreading in PHP.

bshaffer:
Closing as we do not really need this.

PR #810, opened 2016-01-19
automatic token fetching

bshaffer:
see #809

### DIFF
--- a/tests/BaseTest.php
+++ b/tests/BaseTest.php
@@ -61,6 +61,8 @@ class BaseTest extends PHPUnit_Framework_TestCase
 
     $client = new Google_Client();
     $client->setApplicationName('google-api-php-client-tests');
+    $client->setRedirectUri("http://localhost:8000");
+    $client->setConfig('access_type', 'offline');
     $client->setHttpClient($httpClient);
     $client->setScopes([
         "https://www.googleapis.com/auth/plus.me",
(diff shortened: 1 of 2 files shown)


tests/BaseTest.php
```diff
@@ -110,19 +112,41 @@ public function tryToGetAnAccessToken(Google_Client $client)
...
 
-    $client->setRedirectUri("urn:ietf:wg:oauth:2.0:oob");
-    $client->setConfig('access_type', 'offline');
     $authUrl = $client->createAuthUrl();
 
-    echo "\nPlease enter the auth code:\n";
-    ob_flush();
+    $log = tempnam(sys_get_temp_dir(), 'php_auth_log');
+    $cmd = sprintf(
+      'OAUTH_LOG_FILE=%s php -S localhost:8000 -t %s > /dev/null & echo $!',
```
tmatsuo: Does this command work on Windows?

tmatsuo: I'm concerning the envvar

tests/BaseTest.php
```diff
@@ -110,19 +112,41 @@ public function tryToGetAnAccessToken(Google_Client $client)
...
+    if ('1' === $file) {
+      $this->markTestSkipped('Failed to retrieve access token');
+    }
+
+    if (!$accessToken = json_decode($file, true)) {
+      $this->markTestSkipped('invalid JSON: ' . $file);
+    }
+
+    // kill the web server
+    `kill $pid`;
```
tmatsuo: If you hit Ctrl+C while waiting for the token, what happens with the webserver process?

tests/BaseTest.php
bshaffer: it almost certainly does not work on windows, although I haven't checked.

bshaffer:
Yeah I don't think we need this.

PR #812, opened 2016-01-22
Incorrect @return line for Client::authorize

am63:
Was "@return void", whereas the authorize method does return a ClientInferface object in current release.


am63:
I signed it!

bshaffer:
This is great, but "objet" should be "object"

am63:
Typo error has been corrected, sorry !

bshaffer:
merged in c872f606236631bf4da16daa9ff7501d036fed60

Issue #803, opened 2016-01-14
Update DCM API Service to newest version before sunset on Feb 29


tfranke:
The DCM/DFA Reporting and Trafficking API Service (DFAreporting.php) is currently set to work with version v2.1 which will be sunset on Februaray 29 (https://developers.google.com/doubleclick-advertisers/reporting/deprecation).

Will there be a new release that works with the newewst version (v2.3) in time?

mattwhisenhunt:
This issue may no longer be relevant due to its age. Feel free to re-open.

PR #813, opened 2016-01-22
Updating Dfareporting.php

jimper:
Adding an updated DCM API service definition to address #803.

### DIFF
--- a/src/Google/Service/Dfareporting.php
+++ b/src/Google/Service/Dfareporting.php
@@ -16,7 +16,7 @@
  */
 
 /**
- * Service definition for Dfareporting (v2.1).
+ * Service definition for Dfareporting (v2.3).
  *
  * <p>
  * Manage your DoubleClick Campaign Manager ad campaigns and reports.</p>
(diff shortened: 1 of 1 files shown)


bshaffer:
These files are auto-generated, so this change will not take.

noxomatic9000:
Does that mean it won't be updated to support the DCM API v2.3? Or does it mean that the update to the new API version will happen automatically before the sunset date on february 29?

bshaffer:
It means it'll be updated when the script is manually run to update all the APIs. It's not a great solution, but it's all we have right now.

For those who need it sooner can implement the changes in this pull request.

noxomatic9000:
Thanks for the info. Will the API update be available before the sunset date? Also will it be available in the v1 branch?

swordbeta:
@bshaffer _When_ will you manually run the script to update all the APIs? I don't really wanna fork and apply the changes myself.

bshaffer:
these have been updated for `v1-master` (see #834), and the PR for `master` is here: #835 

I won't be merging these changes into `master` because they break BC. I will need to tag a stable release (`2.0.0`) first, and then I can merge them in. This will probably be in a week or two. So if you need this, just pull down my personal fork and use that.

bshaffer:
this has now been merged into the `master` branch

PR #816, opened 2016-01-29
Make cURL proxy header calculation bug workaround configurable

gws:
As discussed in #190, this PR adds a configuration option that is designed to be set by users with old cURL versions but where the vendor has backported a fix for the specific bug that was being worked around (again, see #190 for more detail). In that case, we need to allow users the option of disabling our workaround, since there is no reliable way to detect the presence of the fix at runtime.

### DIFF
--- a/src/Google/Config.php
+++ b/src/Google/Config.php
@@ -58,6 +58,10 @@ class Google_Config
         'Google_IO_Abstract' => array(
           'request_timeout_seconds' => 100,
         ),
+        'Google_IO_Curl' => array(
+          'disable_proxy_workaround' => false,
+          'options' => null,
+        ),
         'Google_Logger_Abstract' => array(
           'level' => 'debug',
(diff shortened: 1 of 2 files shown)


gws:
This is only failing CI because of an unrelated issue - `v1-master`'s autoload function is not conventionally named.

mishan:
:+1: :ship: :it:

Gemmu:
I tested this new configurable feature under RedHat 7.1 and 7.2 (latter has the patched libcurl). I found that using this new flag to disable the workaround, made our code functional again on both of our machines. Please do tell if there's anything more I can help with in order to get this to packagist.org.

bshaffer:
This looks great! Thank you for the work and discussion on this. I have fixed the tests and will merge and tag once they pass.

bshaffer:
@Gemmu  tagged in [1.1.7](https://github.com/google/google-api-php-client/releases/tag/1.1.7). Go forth and Packagist.

PR #817, opened 2016-01-29
Fix #806 - Enable custom umask for cache files

### LINKED ISSUE #806: fopen(/tmp/google-api-php-client/e6/e636...): failed to open stream: Permission denied
Hi,
I've updated my google api php client from RC2 to RC4 and I'm now getting a lot of warning (actually thousands of warning emails in my mailbox...) saying: `fopen(/tmp/google-api-php-client/e6/e63617e9de932b7845fe794b2537c09a): failed to open stream: Permission denied`.

The issue might come from the fact that I'm using the library from fpm with a `www-data` user and from cli in `root` (cron job). But I assume this is a common use case and the library should be able to handle it (by using different cache directories depending of the user for example).

Here is the stacktrace in case you need it:

    /home/gestixi/vendor/google/apiclient/src/Google/Cache/File.php 192
    /home/gestixi/vendor/google/apiclient/src/Google/Cache/File.php 177
    /home/gestixi/vendor/google/apiclient/src/Google/Cache/File.php 99
    /home/gestixi/vendor/google/apiclient/vendor/google/auth/src/CacheTrait.php 65
    /home/gestixi/vendor/google/apiclient/vendor/google/auth/src/Middleware/AuthTokenMiddleware.php 138
    /home/gestixi/vendor/google/apiclient/vendor/google/auth/src/Middleware/AuthTokenMiddleware.php 113
    /home/gestixi/vendor/google/apiclient/vendor/google/auth/src/Middleware/AuthTokenMiddleware.php 114
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/PrepareBodyMiddleware.php 42
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/Middleware.php 30
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/RedirectMiddleware.php 68
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/Middleware.php 57
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/HandlerStack.php 67
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/Client.php 268
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/Client.php 98
    /home/gestixi/vendor/google/apiclient/vendor/guzzlehttp/guzzle/src/Client.php 104
    /home/gestixi/vendor/google/apiclient/vendor/google/auth/src/HttpHandler/Guzzle6HttpHandler.php 33
    /home/gestixi/vendor/google/apiclient/src/Google/Http/REST.php 73
    /home/gestixi/vendor/google/apiclient/src/Google/Task/Runner.php 181
    /home/gestixi/vendor/google/apiclient/src/Google/Http/REST.php 57
    /home/gestixi/vendor/google/apiclient/src/Google/Client.php 758
    /home/gestixi/vendor/google/apiclient/src/Google/Service/Resource.php 232
    /home/gestixi/vendor/google/apiclient/src/Google/Service/Calendar.php 1636

Hope you guys can fixe this issue. Let me know if you need more details.

huebs:
While #617 is well intentioned for its security concerns, it has an annoying side effect when the cache is shared between an application's CLI and Web Interface.  Unless you use the same username for Apache or PHP-FPM to access the CLI (which you shouldn't), a cache file will only be readable by whichever interface created it.  I believe this is what caused the `failed to open stream: Permission denied` error described in #806.

My solution is to add a third constructor argument to `Google_Cache_File` for setting a custom umask.  I've assigned it the default value `077` so that default permissions for files and directories will remain `0600` and `0700` respectively.

### DIFF
--- a/src/Google/Cache/File.php
+++ b/src/Google/Cache/File.php
@@ -30,6 +30,7 @@ class Google_Cache_File implements CacheInterface
 {
   const MAX_LOCK_RETRIES = 10;
   private $path;
+  private $umask;
   private $fh;
 
   /**
@@ -37,10 +38,11 @@ class Google_Cache_File implements CacheInterface
    */
   private $logger;
 
-  public function __construct($path, LoggerInterface $logger = null)
+  public function __construct($path, LoggerInterface $logger = null, $umask = 077)
   {
     $this->path = $path;
     $this->logger = $logger;
+    $this->umask = $umask;
   }
 
   public function get($key, $expiration = false)
@@ -155,7 +157,7 @@ class Google_Cache_File implements CacheInterface
     // trim the directory separator from the path to prevent double separators
     $storageDir = rtrim($this->path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $dirHash;
     if ($forWrite && ! is_dir($storageDir)) {
-      if (! mkdir($storageDir, 0700, true)) {
+      if (! mkdir($storageDir, 0770 & ~$this->umask, true)) {
         $this->log(
             'error',
             'File cache creation failed',
@@ -199,7 +201,7 @@ class Google_Cache_File implements CacheInterface
       return false;
     }
     if ($type == LOCK_EX) {
-      chmod($storageFile, 0600);
+      chmod($storageFile, 0660 & ~$this->umask);
     }
     $count = 0;
     while (!flock($this->fh, $type | LOCK_NB)) {
--- a/tests/Google/CacheTest.php
+++ b/tests/Google/CacheTest.php
@@ -40,6 +40,34 @@ class Google_CacheTest extends BaseTest
     $this->getSetDelete($cache);
   }
 
+  public function testFileWithDefaultMask()
+  {
+    $dir = sys_get_temp_dir() . '/google-api-php-client/tests/';
+    $cache = new Google_Cache_File($dir);
+    $cache->set('foo', 'bar');
+
+    $method = new ReflectionMethod($cache, 'getWriteableCacheFile');
+    $method->setAccessible(true);
+    $filename = $method->invoke($cache, 'foo');
+    $stat = stat($filename);
+
+    $this->assertEquals($stat['mode'] & 0777, 0600);
+  }
+
+  public function testFileWithCustomMask()
+  {
+    $dir = sys_get_temp_dir() . '/google-api-php-client/tests/';
+    $cache = new Google_Cache_File($dir, null, 07);
+    $cache->set('foo', 'bar');
+
+    $method = new ReflectionMethod($cache, 'getWriteableCacheFile');
+    $method->setAccessible(true);
+    $filename = $method->invoke($cache, 'foo');
+    $stat = stat($filename);
+
+    $this->assertEquals($stat['mode'] & 0777, 0660);
+  }
+
   public function testNull()
   {
     $cache = new Google_Cache_Null();


huebs:
One more note about the security of this implementation: it's impossible to grant any permissions to world.  I can't think of any valid reasons for needing that ability.

bshaffer:
Thank you for locating the source of the issue. Your hypothesis sounds pretty believable. 

I am not a huge fan of using `umask` in this way, as I do not believe it's the standard usage. What if we defaulted [`Google\Cache\File::getCacheDir`](https://github.com/google/google-api-php-client/blob/master/src/Google/Cache/File.php#L149) to generate a different hash depending on the user? This way we maintain security by default, but avoid permissions errors for multiple users.

huebs:
I foresee two problems with using hashed usernames as directory names for namespacing.

- CLI cache cleaners will be unable to remove files created by the web interface.
- The web interface will not be able to read files created by a CLI cache warmer.

When creating my pull request, my first inclination was to follow the pattern used by Symfony's [HttpCache][1]:

```php
    /**
     * Save data for the given key.
     *
     * @param string $key  The store key
     * @param string $data The data to store
     *
     * @return bool
     */
    private function save($key, $data)
    {
        $path = $this->getPath($key);
        if (!is_dir(dirname($path)) && false === @mkdir(dirname($path), 0777, true) && !is_dir(dirname($path))) {
            return false;
        }

        $tmpFile = tempnam(dirname($path), basename($path));
        if (false === $fp = @fopen($tmpFile, 'wb')) {
            return false;
        }
        @fwrite($fp, $data);
        @fclose($fp);

        if ($data != file_get_contents($tmpFile)) {
            return false;
        }

        if (false === @rename($tmpFile, $path)) {
            return false;
        }

        @chmod($path, 0666 & ~umask());
    }
```

The same pattern of using php's built-in [`umask()`][2] function is also used by the [`CacheWarmer`][3].  I went with a `$umask` variable rather than a call to `umask()` because I wanted to maintain the library's existing default functionality (directories created with `700` permissions, files saved with `600` permissions).

In the spirit of following standard practices, I've surveyed the current cache implementations of a few PHP frameworks.

The cache implementation (seen above) for [Symfony][4], my framework of choice, creates directories with `777` permissions and there isn't a easy way to customize the behavior.  On the other hand, file permissions can be customized because they are calculated using the default umask: `0666 & ~umask()`.  The [`ClassLoader`][7] creates [directories][5] and [files][6] using the same logic (as do the [`ResourceCheckerConfigCache`][8], [`ContainerBuilderDebugDumpPass`][9], and [`CompilerDebugDumpPass`][10]).

I also looked at [Doctrine][11], Symfony's companion ORM framework, which includes a [`FileLockRegion`][12] cache that creates [directories][13] with `775` permissions by default and [files][14] with `664` permissions -- neither is easily customizable.

Next, I looked at the [Laravel][15] framework. The [`FileStore`][16] creates [directories][17] with `777` permissions using the [`Filesystem::makeDirectory()`][18] method and can't be easily customized, whereas [files][19] are created with system default permissions using the [`Filesystem:put()`][20] method (which simply calls [`file_put_contents()`][21] without a subsequent call to [`chmod()`][22]).

Lastly, I looked at the [CakePHP][23] framework. The cache's [`FileEngine`][24] creates [directories][25] with `775` permissions with no easy method of customization.  Meanwhile, [files][26] are created with permission which are controlled by a configurable [`mask`][27] variable which defaults to [`0664`][28].

**My 2¢:** If the goal is to adhere to standard practices, I'd either use the Symfony (`777` for directories, `0666 & ~umask()` for files) or Laravel (`777` for directories, no change to files) method.

[1]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L333-L364
[2]:  https://secure.php.net/manual/en/function.umask.php
[3]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmer.php#L25
[4]:  https://symfony.com
[5]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/ClassLoader/ClassCollectionLoader.php#L116-L118
[6]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/ClassLoader/ClassCollectionLoader.php#L236-L246
[7]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/ClassLoader/ClassLoader.php
[8]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Config/ResourceCheckerConfigCache.php#L109-L129
[9]:  https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/ContainerBuilderDebugDumpPass.php#L36
[10]: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/Compiler/CompilerDebugDumpPass.php#L29
[11]: http://www.doctrine-project.org/projects/orm.html
[12]: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php
[13]: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php#L64
[14]: https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Cache/Region/FileLockRegion.php#L245
[15]: https://laravel.com
[16]: https://github.com/laravel/framework/blob/5.2/src/Illuminate/Cache/FileStore.php
[17]: https://github.com/laravel/framework/blob/5.2/src/Illuminate/Cache/FileStore.php#L107-L118
[18]: https://github.com/laravel/framework/blob/5.2/src/Illuminate/Filesystem/Filesystem.php#L322-L338
[19]: https://github.com/laravel/framework/blob/5.2/src/Illuminate/Cache/FileStore.php#L90-L105
[20]: https://github.com/laravel/framework/blob/5.2/src/Illuminate/Filesystem/Filesystem.php#L71-L82
[21]: https://secure.php.net/manual/en/function.file-put-contents.php
[22]: https://secure.php.net/manual/en/function.chmod.php
[23]: http://cakephp.org
[24]: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php
[25]: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php#L383
[26]: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php#L391-L406
[27]: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php#L400
[28]: https://github.com/cakephp/cakephp/blob/master/src/Cache/Engine/FileEngine.php#L67

bshaffer:
@huebs if you are willing to make the change, I am all for the symfony approach (777 for directories, 0666 ~ umask() for files). I am not worried about BC, since we haven't tagged a main version yet (being in RC mode allows us to sneak this in)

bshaffer:
closing in favor of #839

PR #823, opened 2016-02-02
Add getData method for easy row data access

thilanga:
Helper method to access data from rows easily and avoid object toSimpleObject casting to an array


thilanga:
I signed it!

PR #825, opened 2016-02-02
Fixed Opening brace issue to fix php 5.4 build issue

thilanga:


bshaffer:
This class gets autogenerated. You could add this method to [`Google\Model`](https://github.com/google/google-api-php-client/blob/master/src/Google/Model.php#L32), which is the subclass where the property comes from. 

Can you explain why you need to access the model data here?

Issue #827, opened 2016-02-02
JWT "BeforeValidException" error in "verifyIdToken"


bshaffer:
The following code occasionally throws a "BeforeValidException" error:

```php
$authResponse = $client->fetchAccessTokenWithAuthCode($code);

if ($client->getAccessToken()) {
    $userInfo = $client->verifyIdToken();
    // .. proceed to do stuff with the user info
}
```

Here is the full stack trace:

```php
BeforeValidException in JWT.php line 112:
Cannot handle token prior to 2016-02-02T13:01:09-0800
in JWT.php line 112
at JWT::decode('eyJhbGciOiJS...', array('RS256')) in Verify.php line 89
at Google_AccessToken_Verify->verifyIdToken('eyJhbGciOiJS...') in Client.php line 682 
```

bshaffer:
fixed in #828

gulachek:
Sign in with Google randomly started throwing this at me.  Can someone add an explanation for what configuration needs to be watched out for either to this issue or to the documentation for verifyIdToken?  Or is this a bug in the google client software?  It's odd that the same code has been working for a year but now is suddenly consistently throwing this message, but as soon as I refresh my sign in with google "data-login_uri" with the same post request that threw the error, the token verification succeeds.  This seems like a bug to me, but I'm open to being incorrect.

I'm happy to provide more context/details, but I want to see if someone is on the other end before putting that effort in.

gulachek:
For future readers: I decided to dig in to the implementation/RFC to figure this out more quickly.

TL;DR The firebase php-jwt JWT implementation has a nonstandard check to verify that "the time according to the JWT on the issuing server" is not later than "the time on the machine that is verifying the JWT", w/in some apparently statically configured leeway.  I believe this check in itself is a bug, sleeping until the 'iat' time in the payload to sync the machine time offsets seems like a sad but doable workaround in the meantime.

More details:

The firebase php-jwt JWT.php has this line:
        // Check that this token has been created before 'now'. This prevents
        // using tokens that have been created for later use (and haven't
        // correctly used the nbf claim).
        if (isset($payload->iat) && $payload->iat > ($timestamp + static::$leeway)) {
            throw new BeforeValidException(
                'Cannot handle token prior to ' . \date(DateTime::ISO8601, $payload->iat)
            );
        }

According to [the RFC](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.6):
[...]

PR #828, opened 2016-02-02
adds 1-second leeway to JWT validation to prevent intermittent except…

bshaffer:
…ions

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -189,10 +189,15 @@ class Google_AccessToken_Verify
 
   private function getJwtService()
   {
+    $jwtClass = 'JWT';
     if (class_exists('\Firebase\JWT\JWT')) {
-      return new \Firebase\JWT\JWT;
+      $jwtClass = 'Firebase\JWT\JWT';
     }
 
-    return new \JWT;
+    // adds 1 second to JWT leeway
+    // @see https://github.com/google/google-api-php-client/issues/827
+    $jwtClass::$leeway = 1;
+
+    return new $jwtClass;
   }
 }


Kak3n:
Should it be fixed in newest version?

bshaffer:
@Kak3n [v2.0.0-RC5](https://github.com/google/google-api-php-client/releases) and on have had this change.

Kak3n:
thanks ok.

2016-05-25 18:49 GMT+02:00 Brent Shaffer <notifications@github.com>:

YuanXiong:
On my server, the system time is 2 min prior to the current time so I got this problem. The system admin may have to run the ntp service for date updates.

jslewis90:
Mine shows 5 hours later than what what the Google Token returns.

bshaffer:
@jslewis90 I imagine this is an issue with your server's date/time configuration.

martinfree:
Now is impossible to set leeway. Please rollback this fix.

bshaffer:
@martinfree You're right, this is not an optimal solution because it pollutes global scope. I'll submit a fix that ensures the adjustment is localized. It would be great if the `Firebase\JWT` class made leeway configurable per instance as well as statically.

martinfree:
@bshaffer thank you for considering my suggestion. Cheers.

hamzawain7:
In google-api-php-client/src/Google/AccessToken/Verify.php line 213 you are not adding 1 to JWT leeway like you mentioned in comments but rather setting it to one. Kindly fix it.
`
if (property_exists($jwtClass, 'leeway')) {
            // adds 1 second to JWT leeway
            // @see https://github.com/google/google-api-php-client/issues/827
            $jwtClass::$leeway = 1;
        }

`

dyegonery:
This solution didn't worked with me. So, I tried adding 60 seconds to JWT leeway:
if (property_exists($jwtClass, 'leeway')) {
// adds 1 second to JWT leeway
// @see #827
$jwtClass::$leeway += 60;
}

FG-Developer:
Hello,

I have that same problem me too, The problem is solving when i changing $leeway variable value as "+=60", but i do not want change in core file of package. But i don't find any way overwrite this variable on another place.

There is any way changing that variable another place?

TheMorganHub:
I'm sorry for the sarcasm that's about to come, but @hamzawain7 is right. The "geniuses" that "fixed" this are not **adding** to the leeway, but instead they're overwriting whatever value you set to it. For now having to manually edit the vendor files will have to do, although a fix would be appreciated. Thanks.

perials:
@FG-Developer You can pass a custom JWT object to `Google_Client`.
```
$jwt = new \Firebase\JWT\JWT;
$jwt::$leeway = 60; // adjust this value

// we explicitly pass jwt object whose leeway is set to 60
$this->client = new \Google_Client(['jwt' => $jwt]);
```
Refer [this](https://perials.com/login-google-using-php-google-php-client/#fix-for-before-valid-exception) post.

PR #839, opened 2016-02-09
File cache umask

bshaffer:
see #806

### DIFF
--- a/src/Google/Cache/File.php
+++ b/src/Google/Cache/File.php
@@ -30,7 +30,6 @@ class Google_Cache_File implements CacheInterface
 {
   const MAX_LOCK_RETRIES = 10;
   private $path;
-  private $umask;
   private $fh;
 
   /**
@@ -38,11 +37,10 @@ class Google_Cache_File implements CacheInterface
    */
   private $logger;
 
-  public function __construct($path, LoggerInterface $logger = null, $umask = 077)
+  public function __construct($path, LoggerInterface $logger = null)
   {
     $this->path = $path;
     $this->logger = $logger;
-    $this->umask = $umask;
   }
 
   public function get($key, $expiration = false)
@@ -157,7 +155,7 @@ class Google_Cache_File implements CacheInterface
     // trim the directory separator from the path to prevent double separators
     $storageDir = rtrim($this->path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $dirHash;
     if ($forWrite && ! is_dir($storageDir)) {
-      if (! mkdir($storageDir, 0770 & ~$this->umask, true)) {
+      if (!mkdir($storageDir, 0777, true)) {
         $this->log(
             'error',
             'File cache creation failed',
@@ -201,7 +199,7 @@ class Google_Cache_File implements CacheInterface
       return false;
     }
     if ($type == LOCK_EX) {
-      chmod($storageFile, 0660 & ~$this->umask);
+      chmod($storageFile, 0666 & ~umask());
     }
     $count = 0;
     while (!flock($this->fh, $type | LOCK_NB)) {
--- a/src/Google/Cache/File.php
+++ b/src/Google/Cache/File.php
@@ -30,6 +30,7 @@ class Google_Cache_File implements CacheInterface
 {
   const MAX_LOCK_RETRIES = 10;
   private $path;
+  private $umask;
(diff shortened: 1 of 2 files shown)


bshaffer:
@huebs please comment here to make the CLA bot happy :D

bshaffer:
@tmatsuo PTAL

src/Google/Cache/File.php
```diff
@@ -155,7 +155,7 @@ private function getCacheDir($file, $forWrite)
     // trim the directory separator from the path to prevent double separators
     $storageDir = rtrim($this->path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $dirHash;
     if ($forWrite && ! is_dir($storageDir)) {
-      if (! mkdir($storageDir, 0700, true)) {
+      if (!mkdir($storageDir, 0777, true)) {
```
tmatsuo: 0777 seems little bit dangerous. What kind of data will be in the cache? Is any sensitive data there? Does it make sense to use 0770 instead?

src/Google/Cache/File.php
```diff
@@ -199,7 +199,7 @@ private function acquireLock($type, $storageFile)
       return false;
     }
     if ($type == LOCK_EX) {
-      chmod($storageFile, 0600);
+      chmod($storageFile, 0666 & ~umask());
```
tmatsuo: `0660` here too

src/Google/Cache/File.php
tmatsuo: Ah nm, it will be umasked anyways, so I think it's fine.

src/Google/Cache/File.php
tmatsuo: nm too

tmatsuo:
LGTM

src/Google/Cache/File.php
bshaffer: Access tokens are stored in cache, yes. But the whole point is to fix #806. So we either break some of our users and be extra secure, or choose to be standard with other file caches and make file permissions looser.

tmatsuo: Ok, let me think about it.

tmatsuo: I think the root user's cache should not be accessible from other users.
Does it make sense to prefix the cache directory with the username?

huebs:
Making bots happy. :smiley:

tmatsuo:
Oh, I LGTMed, but did undo LGTM. Can you revert? See https://github.com/google/google-api-php-client/pull/839#discussion_r52365593

src/Google/Cache/File.php
huebs: @tmatsuo, as I mentioned in #817:

> I foresee two problems with using hashed usernames as directory names for namespacing.
>
> - CLI cache cleaners will be unable to remove files created by the web interface.
> - The web interface will not be able to read files created by a CLI cache warmer.

tmatsuo: > CLI cache cleaners will be unable to remove files created by the web interface.

Why do you need to cleanup?

> The web interface will not be able to read files created by a CLI cache warmer.

Is this a big problem?

Think about this scenario.

I have downloaded the credentials file and stored secretly, only available for root.
root process will create a cache which contains the access token from the credentials.

Then any other users should not be able to read that cache.

The correct approach for the cron scenario is to run the process as an appropriate user, not root.

tests/Google/CacheTest.php
```diff
@@ -40,6 +40,48 @@ public function testFileWithTrailingSlash()
...
+
+    $method = new ReflectionMethod($cache, 'getWriteableCacheFile');
+    $method->setAccessible(true);
+    $filename = $method->invoke($cache, 'foo');
+    $stat = stat($filename);
+
+    $this->assertEquals(0666 & ~umask(), $stat['mode'] & 0777);
+  }
+
+  public function testFileWithCustomMask()
```
huebs: This test case can be removed as it now just duplicates the previous one.

src/Google/Cache/File.php
bshaffer: Added username hash. see #840

bshaffer: Agreed with @tmatsuo this is not a general-purpose file cache. it is a credentials cache. So we'll stay on the safe side here.

PR #840, opened 2016-02-09
use hashed username in cache dir

bshaffer:
Addresses permission concerns with filecache (@tmatsuo PTAL)

### DIFF
--- a/src/Google/Cache/File.php
+++ b/src/Google/Cache/File.php
@@ -153,20 +153,36 @@ class Google_Cache_File implements CacheInterface
     // and thus give some basic amount of scalability
     $fileHash = substr(md5($file), 0, 2);
     $userHash = md5(get_current_user());
-    $dirHash = $fileHash . DIRECTORY_SEPARATOR . $userHash;
+    $dirHash = $userHash . DIRECTORY_SEPARATOR . $fileHash;
 
     // trim the directory separator from the path to prevent double separators
-    $storageDir = rtrim($this->path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . $dirHash;
-    if ($forWrite && ! is_dir($storageDir)) {
+    $rootCacheDir = rtrim($this->path, DIRECTORY_SEPARATOR);
+    $storageDir = $rootCacheDir . DIRECTORY_SEPARATOR . $dirHash;
+
+    if ($forWrite && !is_dir($storageDir)) {
+      // create root dir
+      if (!is_dir($rootCacheDir)) {
+        if (!mkdir($rootCacheDir, 0777, true)) {
+          $this->log(
+              'error',
+              'File cache creation failed',
+              array('dir' => $rootCacheDir)
+          );
+          throw new Google_Cache_Exception("Could not create cache directory: $rootCacheDir");
+        }
+      }
+
+      // create dir for file
       if (!mkdir($storageDir, 0700, true)) {
         $this->log(
             'error',
             'File cache creation failed',
             array('dir' => $storageDir)
         );
-        throw new Google_Cache_Exception("Could not create storage directory: $storageDir");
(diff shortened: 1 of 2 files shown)


src/Google/Cache/File.php
```diff
@@ -151,11 +151,14 @@ private function getCacheDir($file, $forWrite)
     // use the first 2 characters of the hash as a directory prefix
     // this should prevent slowdowns due to huge directory listings
     // and thus give some basic amount of scalability
-    $dirHash = substr(md5($file), 0, 2);
+    $fileHash = substr(md5($file), 0, 2);
+    $userHash = substr(md5(get_current_user()), 0, 2);
```
tmatsuo: Can you just use the whole username? The first 2 letters of the md5 hash of the usernames easily collision.

bshaffer: This is the system user executing the code, so it's pretty unlikely they'll collide.

huebs:
At the very least, the directory permissions need to be relaxed. As this is currently implemented, if the first user to call this function creates any directory defined in the base path (e.g. the default `google-api-php-client` directory), all other users would unable to create their cache directories because they wouldn't have read or write permissions to the base directory. The same goes for the hash prefix directories.

tmatsuo:
@huebs 
That's a good point. I have a proposal.

src/Google/Cache/File.php
```diff
@@ -151,11 +151,14 @@ private function getCacheDir($file, $forWrite)
     // use the first 2 characters of the hash as a directory prefix
     // this should prevent slowdowns due to huge directory listings
     // and thus give some basic amount of scalability
-    $dirHash = substr(md5($file), 0, 2);
+    $fileHash = substr(md5($file), 0, 2);
+    $userHash = md5(get_current_user());
+    $dirHash = $fileHash . DIRECTORY_SEPARATOR . $userHash;
```
tmatsuo: Can you change the order of them?
```php
$dirHash = $userHash . DIRECTORY_SEPARATOR . $fileHash;
```
Then there will be no permission issue.

Also, maybe you can just use the username, instead of a hash of the username? What's the purpose of hashing the username? The md5 hash collision may be unlikely, but from the security standpoint, it needs to be theoretically perfect.

huebs: That doesn't address the problem of the base directory's permissions.  You could solve this by also changing:

```php
if (!mkdir($storageDir, 0700, true)) {
```

to

```php
if (!(mkdir($this->path, 0777, true) && mkdir($storageDir, 0700, true))) {
```

bshaffer: I don't want usernames showing in the directory path, it seems sloppy. Using the full `md5` brings the likelyhood of collisions to essentially zero.

I made the root directory `0777` and the individual ones `0700`. If this isn't acceptable, I'm going to remove this file all together and let everyone write their own implementations (kidding but kinda serious...)

tmatsuo: Ok, I wanted to make sure the collisions are not the problem and I found the actual collision is between binary values way longer than the usual usernames.
```
4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa200a8284bf36e8e4b55b35f427593d849676da0d1555d8360fb5f07fea2
```
collides with
```
4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa202a8284bf36e8e4b55b35f427593d849676da0d1d55d8360fb5f07fea2
```

Thanks for being patient

tmatsuo:
LGTM

bshaffer:
@huebs thoughts on this final iteration?

huebs:
@bshaffer, I see it's already merged but, to answer your question, it looks good to me. Thanks for taking the time to work through this issue (same to @tmatsuo).

Issue #846, opened 2016-02-11
Google_Cache_File use incorrect function 


kunit:
Google_Cache_File use 'get_current_user()', this function does *not* get process user.

http://php.net/manual/en/function.get-current-user.php

Example, test.php owner 'develop'.

```
<?php
printf("get_current_user() : %s\n", get_current_user());
printf("posix function : %s\n", posix_getpwuid(posix_geteuid())['name']);
```

```
develop% php test.php
get_current_user() : develop
posix function : develop
```

```
develop% sudo php test.php
get_current_user() : develop
posix function : root
```

PR #847, opened 2016-02-11
use correct get process user function

kunit:
see #846

### DIFF
--- a/src/Google/Cache/File.php
+++ b/src/Google/Cache/File.php
@@ -154,11 +154,11 @@ class Google_Cache_File implements CacheInterface
     $fileHash = substr(md5($file), 0, 2);
     $processUser = null;
     if (function_exists('posix_geteuid')) {
-        $processUser = posix_getpwuid(posix_geteuid())['name'];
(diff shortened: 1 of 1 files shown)


bshaffer:
Hmm... throwing an exception is worse than the problems this is trying to solve. If `get_current_user` returns nothing, we can `md5` the empty string.

Did you run into an error where you used `sudo` and then later couldn't access the files as a web user?

kunit:
Test Program(test.php) owner 'develop'.

```
require_once '/path/to/your-project/vendor/autoload.php';

umask(0);
$cache = new Google_Cache_File(__DIR__ . '/test');
$cache->set('test', '123');
```

run script use original Google_Cache_File, raise error.
same name directory create, because get_current_user() returns same value 'develop'.

```
develop% sudo php test.php
develop% php test.php
PHP Warning:  mkdir(): Permission denied in /path/to/your-project/src/Google/Cache/File.php on line 176
```

bshaffer:
great! thank you for that. If you remove the exception, I will merge this.

kunit:
remove the exception, change log-level.

PR #854, opened 2016-02-16
echo missingOAuth2CredentialsWarning

itsazzad:
missingOAuth2CredentialsWarning should `echo` and `exit`;


bshaffer:
this is a good find, but we don't want `exit` in our tests. I have fixed this as part of #835

bshaffer:
this has now been merged into the `master` branch

PR #857, opened 2016-02-18
removed unused line

itsazzad:
```
$service = new Google_Service_Drive($client);
```
This line is not seems to be necessary here.


daniroyo:
Agree

PR #858, opened 2016-02-19
id_token_token instead of id_token

itsazzad:
Should it not be so?


bshaffer:
looks good! Thank you

PR #862, opened 2016-02-22
Fix listTranslations to make getTranslations work

PurpleBooth:
If you make a listTranslations call you'll get a ```Google_Service_Translate_TranslationsListResponse``` back. This has a data value which is a key with "translations" array in it. 

Unfortunately these aren't returned by the getTranslations call,  you must instead use ```$translations->data['translations']```. This fixes that method so it works.

### DIFF
--- a/src/Google/Service/Translate.php
+++ b/src/Google/Service/Translate.php
@@ -334,11 +334,11 @@ class Google_Service_Translate_TranslationsListResponse extends Google_Collectio
 
   public function setTranslations($translations)
   {
-    $this->translations = $translations;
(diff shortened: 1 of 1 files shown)


PurpleBooth:
I signed it!

PurpleBooth:
Oh, actually signed it this time. Oops!

PurpleBooth:
I signed it!

PurpleBooth:
I signed it as `Purple Booth Ltd`

PurpleBooth:
Looks like this is actually from #463. As this is a generated library this won't fix it.

PR #873, opened 2016-02-25
for raw response: fix alt=media detection

Fneufneu:
getUri()->getQuery("alt") return `alt=media` not `media`
$isJson is correct with this change

### DIFF
--- a/src/Google/Http/REST.php
+++ b/src/Google/Http/REST.php
@@ -106,7 +106,7 @@ class Google_Http_REST
     $result = null;
 
     // return raw response when "alt" is "media"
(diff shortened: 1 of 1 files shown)


Fneufneu:
I signed it!

Fneufneu:
I signed it!

bshaffer:
@Fneufneu Good catch that this will not work.  However, we can make this more robust/accurate:

```php
$isJson = true;
if ($request && $qs = $request->getUri()->getQuery()) {
    parse_str($qs, $query);
    if (isset($query['alt']) && $query['alt'] == 'media') {
        $isJson = false;
    }
}
```

bshaffer:
closing in favor of #876

PR #874, opened 2016-02-25
Fix large file download

Fneufneu:
decodeHttpResponse() store the body response in a variable so if you try to download a large file from you Drive, it will crash with out of memory error.
Since the data is not needed for raw response, do it only when we try to json_decode().

I'm not sure of the modification since $body is used in Google_Service_Exception() in case of retry.
But with this change and #873 , i'm able to backup my Drive successfully.


bshaffer:
Thank you for finding this issue!

bshaffer:
closing in favor of #876

PR #876, opened 2016-02-26
cleans up REST::decodeResponse

bshaffer:
See #874  and #873

### DIFF
--- a/src/Google/Http/REST.php
+++ b/src/Google/Http/REST.php
@@ -107,44 +107,76 @@ class Google_Http_REST
       RequestInterface $request = null,
       $expectedClass = null
   ) {
-    $body = (string) $response->getBody();
     $code = $response->getStatusCode();
-    $result = null;
-
-    // return raw response when "alt" is "media"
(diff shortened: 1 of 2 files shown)


src/Google/Http/REST.php
```diff
@@ -107,44 +107,70 @@ public static function decodeHttpResponse(
...
-      if (isset($result['error']['errors'])) {
-        $errors = $result['error']['errors'];
-      }
-      throw new Google_Service_Exception($body, $code, null, $errors);
+    return $response;
+  }
+
+  private static function decodeBody(ResponseInterface $response, RequestInterface $request = null)
+  {
+    if (self::isAltMedia($request)) {
```
tmatsuo: It might be an edge case, but if the isAltMedia == true, and the response was an error, it seems that the error message will be thrown away.

bshaffer: This is true. We can probably assume if it's an error, it's safe to pull the body into memory.

bshaffer: fixed! PTAL

tmatsuo:
LGTM

PR #878, opened 2016-02-29
Check if expires_in is set in token.

limonte:
This PR solves: 

```
ErrorException in Client.php line 462:
Undefined index: expires_in

in Client.php line 462
at HandleExceptions->handleError('8', 'Undefined index: expires_in', '/home/limon/www/youtube-search/vendor/google/apiclient/src/Google/Client.php', '462', array('created' => '0')) in Client.php line 462
at Google_Client->isAccessTokenExpired() in Client.php line 351
at Google_Client->authorize() in Client.php line 756
at Google_Client->execute(object(Request), 'Google_Service_YouTube_PlaylistListResponse') in Resource.php line 232
at Google_Service_Resource->call('list', array(array('part' => 'id', 'mine' => 'true')), 'Google_Service_YouTube_PlaylistListResponse') in YouTube.php line 3705
at Google_Service_YouTube_Playlists_Resource->listPlaylists('id', array('mine' => 'true')) in AuthController.php line 76
```

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -458,10 +458,14 @@ class Google_Client
     }
 
     // If the token is set to expire in the next 30 seconds.
-    $expired = ($created
-      + ($this->token['expires_in'] - 30)) < time();
+    if (isset($this->token['expires_in'])) {
+        $expired = ($created
+          + ($this->token['expires_in'] - 30)) < time();
 
-    return $expired;
+        return $expired;
+    }
+
+    return false;
   }
 
   public function getAuth()


limonte:
I signed it!

limonte:
There's an issue in CI build, I need some help:

zaeyct0

bshaffer:
This changes the behavior of the function. If `expires_in` *isn't set*, the function returns `false` (NOT expired), when previously this would return `true` (expired), as `expires_in` would be evaluated as zero.

Are you intentionally trying to change the behavior, or are you just trying to remove the `undefined index` warning? If you are doing the latter, this would be better:

```php
$expiresIn = isset($token['expires_in']) ? $token['expires_in'] : 0;
$expired = ($created
    + ($this->token['expires_in'] - 30)) < time();
  		  
return $expired;
```

limonte:
Thank @bshaffer for the explanation. I think I'm fixing my error in wrong place.

This might be the right place: https://github.com/laravel/socialite/pull/135

PR #893, opened 2016-03-14
switches this library to pull in generated classes via composer

bshaffer:
see https://github.com/google/google-api-php-client-services


tmatsuo:
LGTM

Issue #901, opened 2016-03-18
How do Refresh Token results work, if done automatically within the API?


jonnywilliamson:
Hi, I've no problem with understanding the refresh token, it's purpose, how to get it OR how to use it. 

I am confused however about how the api uses it automatically when an access_token has expired. Here's a short example.

My Code:

1. User approves access (with offline access)
2. Refresh token and access token returned with time to expire (for access token)
3. Refresh token saved to database for use.
4. 1 day later, code retrieves `access_token` and expiry time from database. Detects that token should be expired. (uses `isAccessTokenExpired()`)
5. *Manually* use refresh token to get new `access_token`. Both saved/updated in database.
6. New `access_token` used to get data from google service.

That's all fine.

However, I see in the code in `Google_Client` that the check to see if the `access_token` is expired is already called when the `authorise` method is called AND if it is expired and the refresh token has been supplied, it will use the refresh token when preparing the credentials.

This is where I get confused. 

At no stage during the execution of the `Google_Client` code, can I find a place where the results of using the refresh_token can be 'intercepted' and thus, saved/updated to the users database.

I'd love to be able to remove all the time checking and `isAccessTokenExpired()` code from MY code and let the API deal with it (as it currently is). However, I DO want to know, if the API uses the refresh_token, what is the new `access_token` and when does it expire! I want to be able to update my database so that if I send a second request a few seconds later, it doesn't use the OLD `access_token` like it currently does in my codebase.

Can anyone shed somelight on what/how this works?

Many thanks.

bshaffer:
As long as the `refresh_token` key is set in the array passed to [`Google_Client::setAccessToken`](https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L399), the access token should be refreshed automatically.

If you're looking to save the access token somewhere, this *is* tricky, as it's not currently being saved back to the `Google_Client`. But it is stored in cache, so you could retrieve it using `$client->getCache()`.

jonnywilliamson:
>If you're looking to save the access token somewhere, this is tricky, as it's not currently being saved back to the Google_Client. But it is stored in cache, so you could retrieve it using $client->getCache().

Yes just so there's no confusion to my question:

I can see that the API is using the `refresh_token` automatically for me - also my request for data is successful so I know the API is working well. As you say, I just can't see how to get my hands on the new `access_token` and `expires_in` values etc from the request the API sent.

[...]

Issue #849, opened 2016-02-11
Issue with Google_Client no longer using caching with subsequent API calls


am63:
Hi,

Since this commit, most API calls are broken in my application.

Here are some details about the issue and when it occurs :

- single API call (such as a new event in Calendar) : OK, no issue
- 2 or more API calls, with a recent login (less than one hour) ; for example, read an event in Calendar, and send back the event with some updates : OK, no issue
- **2 or more API calls, with a login done more than one hour ago : KO**, exception is thrown. If I delete the token (which I store in a database), the user logs again, and all is OK.

The error returned by the API has the following data :
`'Client error: POST https://www.googleapis.com/oauth2/v3/token resulted in a 400 Bad Request response: { "error": "invalid_request", "error_description": "Required parameter is missing: grant_type" } '`

and here is the stack trace :

```
#0 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/guzzle/src/Middleware.php(65): GuzzleHttp\Exception\RequestException::create(Object(GuzzleHttp\Psr7\Request), Object(GuzzleHttp\Psr7\Response))
#1 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/promises/src/Promise.php(199): GuzzleHttp\Middleware::GuzzleHttp\{closure}(Object(GuzzleHttp\Psr7\Response))
#2 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/promises/src/Promise.php(152): GuzzleHttp\Promise\Promise::callHandler(1, Object(GuzzleHttp\Psr7\Response), Array)
#3 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/promises/src/TaskQueue.php(60): GuzzleHttp\Promise\Promise::GuzzleHttp\Promise\{closure}()
#4 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/promises/src/Promise.php(240): GuzzleHttp\Promise\TaskQueue->run(true)
#5 /home/assistano/www/lib/google-api-php-client-2.0.0-RC5/vendor/guzzlehttp/promises/src/Promise.php(217): GuzzleHttp\Promise\Promise->invokeWaitFn()
[...]

PR #905, opened 2016-03-22
provides way to receive last token

bshaffer:
see #901

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -75,6 +75,11 @@ class Google_Client
   private $logger;
 
   /**
+   * @var Google\Auth\FetchAuthTokenInterface
+   */
+  private $lastCredentials;
+
+  /**
    * @var boolean $deferExecution
    */
   private $deferExecution = false;
@@ -360,6 +365,7 @@ class Google_Client
 
     if ($credentials) {
       $http = $authHandler->attachCredentials($http, $credentials);
+      $this->lastCredentials = $credentials;
     } elseif ($token) {
       $http = $authHandler->attachToken($http, $token, (array) $scopes);
     } elseif ($key = $this->config['developer_key']) {
@@ -985,6 +991,13 @@ class Google_Client
     }
 
     return $this->http;
+  }
+
+  public function getLastReceivedToken()
+  {
+    if ($this->lastCredentials) {
+      return $this->lastCredentials->getLastReceivedToken();
+    }
   }
 
   protected function createDefaultHttpClient()
--- a/tests/Google/ClientTest.php
+++ b/tests/Google/ClientTest.php
@@ -403,6 +403,29 @@ class Google_ClientTest extends BaseTest
   }
 
   /**
+   * Test after authentication the last received token can be retrieved
+   */
+  public function testGetLastReceivedToken()
+  {
+    $this->checkServiceAccountCredentials();
+
+    // create a NEW client
+    $client = new Google_Client();
+    $client->setScopes(["https://www.googleapis.com/auth/plus.me"]);
+    $client->setAuthConfig(getenv('GOOGLE_APPLICATION_CREDENTIALS'));
+    $this->assertNull($client->getLastReceivedToken());
+
+    $http = $client->authorize();
+    $response = $http->get('https://www.googleapis.com/plus/v1/people/me');
+
+    $token = $client->getLastReceivedToken();
+
+    $this->assertNotNull($token);
+    $this->assertArrayHasKey('access_token', $token);
+    $this->assertArrayHasKey('expires_at', $token);
+  }
+
+  /**
    * Test that the ID token is properly refreshed.
    */
   public function testRefreshTokenSetsValues()


bshaffer:
@tmatsuo PTAL

jonnywilliamson:
@bshaffer 

So I took a look at this last night, it works great! 

I can easily get access to the `access_token` and `expiry` time now if the API is forced to refresh the `access_token` because the one I supplied has expired etc.

However, I'm left wondering how best to use this. As far as I can see there's no event system (or any observation pattern) in the API, so I'm struggling to come up for the best way to know when to call the `getLastReceivedToken` method so I can save/update the new `access_token`

When I make more than one request, the `getLastReceivedToken` method always seems to return `NULL`. I'm not sure why...

Here's some pseudo code of what I mean:

Imagine my saved `access_token` in the database is now a day old...
```php

    $client = new Google_Client(['myconfigvalues']);
    $client->setAccessToken('OldaccessTokenRetrievedFromDatabase');

    $cal = new Google_Service_Calendar($client);

    //This will show the expired token that was taken from the database
    var_dump($client->getAccessToken());


    //We make a request and the API magically asks and receives a NEW access_token. Brilliant.
    $request1 =  $cal->calendarList->listCalendarList();
    var_dump($client->getLastReceivedToken()); //New access_token

   //So far so good. I could technically save this data right now. But say I wanted to do a number of
   //requests....checking after EVERY request is very tedious.

    //Let's fire off another request:
    $request2 = $cal->calendarList->get('myemailaddress@gmail.com');
    var_dump($client->getLastReceivedToken());  //Null returned

    //MMMm. Now getLastReceivedToken() is showing NULL
```

I'm sure this is something to do with the cache, but I'm not sure.

As it stands this is a great start. However, my work flow has me making multiple requests to the google_service rather than just one. Unless I'm coded something badly, it appears that the `getLastReceivedToken` method only works after the 1st request.

What I was hoping to do was override the `__destruct` method on the `Google_Client` and when it was being destroyed I was going to check for any new tokens and save them to the database for that user for the next time.

Have I picked this up wrong?

jonnywilliamson:
So tracing through the code, I think this is where the issue 

So it all works well for the 1st request, but if there is a second request, in `Google_Client` the `authorize` method gets called and this code is used:

image

On line 353, `$this->getAccessToken` is called again, but because the `Google_client` has never been updated with the new `access_token` that it managed to get in the last request, it still retrieves the old stale one that was set when `Google_Client` was created.

That means that the `lastCredentials` gets reset to just the refresh token...wiping away the previous new `access_token`.

Now in the `AuthTokenMiddleware`:

image

The cache gets set, and uses the last new `access_token` to make the second request which will work perfectly.

However at no stage is this passed back to the `Google_Client`.


I have no idea what to do to fix that, I'm struggling getting my head around the architecture here, but I hope that helps to explain what the issue is.

Thank you!

am63:
Hi,

I had similar issue as yours (saving and refreshing the token). Why do you bother with detecting a refreshed token, whereas you could just ask the library whether the token is expired or not ? If so, you commit it to your database (for future use). If not, you continue to use it. 

Here is a piece of code I'm using :

```
// read token from the API or from your database, if it has been saved
$t = $client->getAccessToken();
print_r($t); // should output several fields (including expires_in and refresh_token)
			
// if token is expired
if ( $client->isAccessTokenExpired() )
{
   // fetch refreshed token, but this one is short-lived (one hour validity) ; you CAN'T save it to your 
   // database for future use !
   $short_lived_token = $client->fetchAccessTokenWithRefreshToken($t['refresh_token']);

   // for future calls in the API and the script lifetime, set the refresh token just obtained
   $client->setAccessToken($short_lived_token);
			
   $t['access_token'] = $short_lived_token['access_token'];
   $t['created'] = $short_lived_token['created'];
   $t['expires_in'] = $short_lived_token['expires_in'];

   // do some stuff to save the token in your database, for future use
}
```

Hope this helps a bit !

am63:
Are you sure you are using the last RC ? As your issue reminds me of mine described here (and addressed by a pull request)

#https://github.com/google/google-api-php-client/issues/849

jonnywilliamson:
HI @am63 Thanks for replying.

I'm definitely using RC6. From my composer.lock file:
image

I'm currently using something very similar to your code at the moment, testing the access_token when I set it with the Users saved one.

If it's expired/stale I get a new one and save it, then set that one in the `Google_Client`.

However all that work is done already in the API! It checks if it's expired. It checks if its invalid. It fetches a new one using the refresh token. It uses the new one for any requests (via the cache it seems).

Just trying to see why we need to write that boilerplate code to do all that when all I need is to grab the tokens that were used by the library when it's shutting down.


There's one line in your code that surprises me:

>// fetch refreshed token, but this one is short-lived (one hour validity) ; you CAN'T save it to your 
>   // database for future use !

Why do you say that? 

I have plenty of requests that will be done within an hour of the last one, and so using the token again is fine. Just curious.

Thanks though!

am63:
I say that because the token obtained by fetchAccessTokenWithRefreshToken does not contain any refresh_token field ; the expires_in field is set to expire the token 1 hour later. If you save the fetchAccessTokenWithRefreshToken in the database, you are loosing the refresh_token part !

This is what I learnt some weeks ago, I don't think it has changed.

jonnywilliamson:
@am63 Ah ok. 

I'm working with Laravel, luckily it's smart enough to know that if I send an update command to the database with an array that only has an `access_token` and `expires_in` field, it will update only those fields and not remove the `refresh_token` etc for that user.

But as a warning for others, that's fair enough :)

jonnywilliamson:
I'm sorry for high-jacking this PR conversation, but things just keep jumping out at me.

So @am63  you said earlier on:
>Why do you bother with detecting a refreshed token, whereas you could just ask the library whether the token is expired or not ? 

And you followed it up with this code:
```php
$t = $client->getAccessToken();
// if token is expired
if ( $client->isAccessTokenExpired() ){
[...]
}
```

This is actually pretty worthless IMHO.

When we have a look at the `isAccessTokenExpired()` method, you find that it's nothing more that a time checker.

The code looks like this (comments removed):
```php
  public function isAccessTokenExpired()
  {
    if (!$this->token) {
      return true;
    }

    $created = 0;
    if (isset($this->token['created'])) {
      $created = $this->token['created'];
    } elseif (isset($this->token['id_token'])) {
      $idToken = $this->token['id_token'];
      if (substr_count($idToken, '.') == 2) {
        $parts = explode('.', $idToken);
        $payload = json_decode(base64_decode($parts[1]), true);
        if ($payload && isset($payload['iat'])) {
          $created = $payload['iat'];
        }
      }
    }
    // If the token is set to expire in the next 30 seconds.
    $expired = ($created
            + ($this->token['expires_in'] - 30)) < time();
    return $expired;
  }
```

As you can see, it literally only checks to see if the time left on the token is more than 30 secs otherwise it just says the token has expired.

This does no checking to see if the token has been revoked, if it's actually a valid token (ie, the string of characters hasn't be tampered with)  etc.

If you really wanted to check is the token valid, you need to check with Google. Something like this:
```php
    protected function isAccessTokenInvalid($token)
    {
        $http = $this->client->authorize();

        $result = json_decode(
            $http->get('https://www.googleapis.com/oauth2/v1/tokeninfo?access_token='.$token)->getBody()
        );

        if (isset($result->error)) {
            //Bad Token - invalid, corrupt etc.
            return true;
        }

        return false;
    }
```


Basically it seems the work flow should be simple:

1) Create Google_Client and add a token. (It might be old and invalid, it might be new and valid).
2) Make requests. Api detects what needs to happen if token is old. Gets new token and uses it. (Currently the API does this).
3) When finished all requests, return the tokens used. They can be saved if needed for the next time.

Wouldn't that bypass all the checks and bolierplate code we have to do at the moment?

tmatsuo:
Guys thanks for the discussion.

@bshaffer Does it make sense to create the UserRefreshCredentials only when
- the original access token is expired
- isset($token['refresh_token'])
- the credential is not UserRefreshCredentials
?

bshaffer:
@tmatsuo yes, but that is the current behavior. I'm not quite sure what you mean by the last point, though.

tmatsuo:
Now the if clause is

```php
      if ($this->isAccessTokenExpired() && isset($token['refresh_token'])) {
        $credentials = $this->createUserRefreshCredentials(
            $scopes,
            $token['refresh_token']
        );
      }
```

It causes the $credentials will be cleared when you call the 2nd API call.

So I'm thinking something like:

```ph
      if ($this->isAccessTokenExpired() && isset($token['refresh_token'])
          && ! $credentials instanceof UserRefreshCredentials) {
        $credentials = $this->createUserRefreshCredentials(
            $scopes,
            $token['refresh_token']
        );
      }
```

so that if you call the API again, the $credentials will keep the same object. I think this will solve the issue described in https://github.com/google/google-api-php-client/pull/905#issuecomment-202046445

bshaffer:
@tmatsuo the code never allows for [`$credentials`](https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L352) to be set at that point. The only other place it's set is at [line e47](https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L347).

tmatsuo:
@bshaffer Oh I got it. Somehow I thought $credentials is an instance variable.

Well, then how about to have an instance variable that holds the UserRefreshCredentials and do the same check?

```php
    /**
     * @var UserRefreshCredentials $userRefreshCredentials
     */
    private $userRefreshCredentials;
    ...
    ...
    if ($this->isAccessTokenExpired() && isset($token['refresh_token'])) {
        if (!isset($this->userRefreshCredentials)) {
            $this->userRefreshCredentials = $this->createUserRefreshCredentials(
                $scopes,
                $token['refresh_token']
            );
        }
        $credentials = $this->userRefreshCredentials;
    }
```

tmatsuo:
Oh, we already has $this->lastCredentials, then we can check

```
    if ($this->isAccessTokenExpired() && isset($token['refresh_token'])) {
        if (! $this->lastCredentials instanceof UserRefreshCredentials) {
            $credentials = $this->createUserRefreshCredentials(
                $scopes,
                $token['refresh_token']
            );
        } else {
          $credentials = $this->lastCredentials;
        }
    }
```
What do you think?

bshaffer:
Here are two options which may provide a better solution to this problem:

1. The `Google_Client` class could provide a callback function to `AuthHandlerFactory`, which would be called when a brand new access token is fetched. We would need to add support to set a callback function in the [`AuthTokenSubscriber`](https://github.com/google/google-auth-library-php/blob/master/src/Subscriber/AuthTokenSubscriber.php) and [`AuthTokenMiddleware`](https://github.com/google/google-auth-library-php/blob/master/src/Middleware/AuthTokenMiddleware.php) classes.

2. A `CacheInterface` class called `CallbackCache` could be added. The callback cache essentially takes a callback function and an optional `CacheInterface` class as constructor arguments, and calls both of them when a new value is set. The `auth` library would not require modification for this.

tmatsuo:
Sure, if we can set the brand new access token to the client object, that's better.

tmatsuo:
I'm fine with both of them

jonnywilliamson:
I would love to be able to say which of those two options seems better, but I'm not sure I'm up to speed with the entire API code base to give a constructive opinion.

I am happy to give way to your better judgement.

Thank you for reading though my long rants above :)

bshaffer:
Does anyone on this thread think implementing PSR-6 (#943), and in doing so supporting [stash](https://github.com/tedious/Stash), is sufficient to solve this use-case? I can add something like `setTokenCallback` as a catch-all for any usecase (similar to #925), but this seems hackish and I am wondering if exposing this using a cache interface isn't the better way to handle it.  

All thoughts are welcome!

jonnywilliamson:
Again apologies, but the question is above my skill level to answer!

Issue #916, opened 2016-04-06
ExpiredException in src/Google/AccessToken/Verify.php doesn't resolve


danielmirchandani:
See https://github.com/google/google-api-php-client/blob/v2.0.0-RC7/src/Google/AccessToken/Verify.php#L108. I'm under the impression that:

use Firebase\JWT\ExpiredException;

is required at the top of the file in order for ExpiredException to resolve to the desired class. It looks like a few other class references in that file need "use" statements to match.

danielmirchandani:
I realize now that the rest of the classes are in the global namespace, so "ExpiredException" is the only class that needs an additional "use" statement.

danielmirchandani:
This was solved with pull-request #919.

PR #919, opened 2016-04-07
ExpiredException didn't have the correct namespace

danielmirchandani:
See issue #916.

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -16,6 +16,7 @@
  * limitations under the License.
  */
 
+use Firebase\JWT\ExpiredException;
 use Google\Auth\CacheInterface;
 use GuzzleHttp\Client;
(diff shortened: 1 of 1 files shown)


bshaffer:
Unfortunately this is not that simple. as this library supports `v2.0` of firebase as well, which [does not use the namespace](https://github.com/firebase/php-jwt/blob/v2.2.0/Exceptions/ExpiredException.php)

bshaffer:
It would be best to do the following:

```php
use Firebase\JWT\ExpiredException as ExpiredExceptionV3;

//and then below

try {
    //...
} catch (ExpiredException $e) {
    //...
} catch (ExpiredExceptionV3 $e) {
    //...
}
```

danielmirchandani:
I've updated the pull request based on your comments.

bshaffer:
added in bb90330039782d1dbcfd8a560740efd032ee4283

Thanks!

PR #936, opened 2016-04-28
Update composer.json

enumag:
There should be no problem with guzzlehttp/psr7 version 1.3.


enumag:
Signed.

bshaffer:
This same change will be added in #943

PR #943, opened 2016-05-05
adds PSR-6 and removes custom caching

bshaffer:
dependent on the merging and tagging of https://github.com/google/google-auth-library-php/pull/108


bshaffer:
@tmatsuo @SurferJeffAtGoogle PTAL

tmatsuo:
One question. Looking briefly, I think now it's a user's responsibility to configure the cache. Am I right? If so, can you add a brief documentation too?

bshaffer:
@tmatsuo - yes that's correct. A user needs to install a psr-6 compatible caching library like Stash. Additionally they can supply an anonymous function as a "token callback" which will get called whenever a new token is fetched. 

I will add all this to the README.

bshaffer:
@tmatsuo PTAL!

tmatsuo:
LGTM

One question. Can you guess the cache key? If so, I think you can always get the current auth token from the cache. Do we still need the callback?

bshaffer:
@tmatsuo you *can* guess the cache key, but there's no easy way to expose this. Also, being able to query the cache and see if a token is changed isn't as clean as providing a callback.

I agree the callback isn't strictly necessary, but i do think it's a very nice feature for some use cases (see #901)

tmatsuo:
LGTM

PR #952, opened 2016-05-18
uses custom autoloader instead of classmap - much better

bshaffer:
Classmaps are the anti-autoloader pattern. With the number of classes this pulls in, this is memory-intensive (every class-lookup pulls in the classmap).

We have a wonderful custom autoloader in `v1` that we should just tell composer to use. This will make everyone happy.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,9 +13,7 @@
         "squizlabs/php_codesniffer": "~2.3"
     },
     "autoload": {
-        "classmap": [
-            "src/"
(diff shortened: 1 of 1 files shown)


bshaffer:
This also allows anyone using `v1-master` to use the latest google services simply by requiring the composer package:

```php
composer install google/apiclient-services
```

And the autoloader will load these services before executing the custom autoloader! woohoo.

PR #959, opened 2016-05-25
Add support for new "whoCanAdd" method

usaussie:
This adds support for the new "whoCanAdd" method, as documented on: https://developers.google.com/admin-sdk/groups-settings/v1/reference/groups#resource-representations

### DIFF
--- a/src/Google/Service/Groupssettings.php
+++ b/src/Google/Service/Groupssettings.php
@@ -177,6 +177,7 @@ class Google_Service_Groupssettings_Groups extends Google_Model
   public $sendMessageDenyNotification;
   public $showInGroupDirectory;
   public $spamModerationLevel;
+  public $whoCanAdd;
   public $whoCanContactOwner;
   public $whoCanInvite;
   public $whoCanJoin;
@@ -353,6 +354,14 @@ class Google_Service_Groupssettings_Groups extends Google_Model
   public function getSpamModerationLevel()
   {
     return $this->spamModerationLevel;
+  }
+  public function setWhoCanAdd($whoCanAdd)
+  {
+    $this->whoCanAdd = $whoCanAdd;
+  }
+  public function getWhoAdd()
+  {
+    return $this->whoCanAdd;
   }
   public function setWhoCanContactOwner($whoCanContactOwner)
   {


usaussie:
I signed it!

bshaffer:
These files are autogenerated, so this change would get overwritten. 

If you're wanting to use this updated models in `v1`, follow the [instructions here](https://github.com/google/google-api-php-client-services#usage-in-v1) and let me know if you have any problems.

usaussie:
Thanks Brett. I included google/apiclient-services:dev-master per your note, but my app still makes use of the V1 client library, and is using the groupssettings class from it (instead of from the apiclient-services package). Is there a best practice method of forcing use of the newer class, seeing as though they share the same namespace?

bshaffer:
How are you managing autoloading? Are you including `vendor/autoload.php` at the top of your PHP file?

usaussie:
It's part of a Zend Framework app, and yes, the autoload.php is included at
the top of the main index.php file.

......................................
*Nick Young*
usaussie@gmail.com

bshaffer:
Ahh, I think I know the issue. Try using the version `dev-v1-master`:

```
composer require google/apiclient:dev-v1-master
```

or in composer.json:

```json
{
    "require": {
        "google/apiclient": "dev-v1-master"
    }
}
```

usaussie:
I think that did it. It's working in my local environment. Now i just need to get an updated version of PHP in my production environment... as i was welcomed with this during composer update :-(

`google/apiclient-services dev-master requires php >=5.4 -> your PHP version (5.3.3)`

So.... changing my initial composer.json from

```
 "google/apiclient": "1.x",
```
to 
```
"google/apiclient-services": "dev-master",
"google/apiclient": "dev-v1-master",
```

seems to have done the trick.

Thanks for the help.

bshaffer:
👍 great!

bshaffer:
For the record, you can install with composer using the flag `--ignore-platform-reqs`, and you'll be able to get around the PHP version issue. Also, I can update `google/apiclient-services` to run on PHP 5.3, as I don't think 5.4 is actually required.

usaussie:
`--ignore-platform-reqs`

I've learned my one thing for the day. Time to head home :-)

Thanks again.

PR #972, opened 2016-06-09
adds service account info to README

bshaffer:


README.md
```diff
@@ -77,6 +79,31 @@ foreach ($results as $item) {
...
+### Authentication with Service Accounts ###
+
+> An example of this in action can be seen in [`examples/service-account.php`](examples/service-account.php).
+
+1. Follow the instructions to [Create a Service Account](https://developers.google.com/api-client-library/php/auth/service-accounts#creatinganaccount)
+1. Download the JSON credentials
+1. Set the path to these credentials using the `GOOGLE_APPLICATION_CREDENTIALS` environment variable:
+
+    ```php
+    putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/service-account.json');
```
tmatsuo: Nit: Embedding the environment variable into the code will reduce the benefit from Application Default Credentials.
I think there is ADC support, right?

bshaffer: I could add a note that this is only necessary outside of GAE/GCE, in addition to adding a "for example" above, so it's clear `putenv` is just one way to set an environment variable.

bshaffer: a note has been added

SurferJeffAtGoogle:
LGTM.

PR #973, opened 2016-06-09
Stablize dependency versions

AwesomeBobX64:
When attempting to install the google/apiclient library at two different times, breaking changes were introduced in on of the google/apiclient's dependencies (google/apiclient-services).

This is due to the fact that a .gitignore file is not versioned controlled, which would allow locking down only stable versions of dependencies.

Additionally, the dependency that was updated (google/apiclient-services) is referenced by `@dev` which allows going from v0.2 to v0.3.

There are breaking changes in the v0.3 of google/apiclient-services, namely class name changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .DS_Store
 phpunit.xml
-composer.lock
 vendor
 examples/testfile-small.txt
 examples/testfile.txt
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "require": {
         "php": ">=5.4",
         "google/auth": "0.8",
-        "google/apiclient-services": "*@dev",
+        "google/apiclient-services": "v0.2",
         "firebase/php-jwt": "~2.0|~3.0",
         "monolog/monolog": "^1.17",
         "phpseclib/phpseclib": "~2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,2030 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "This file is @generated automatically"
+    ],
+    "hash": "26049d940d04477f4f7354b99dfe6d0e",
+    "content-hash": "b699a58449f754171d549c28b9492e16",
+    "packages": [
+        {
+            "name": "firebase/php-jwt",
+            "version": "v3.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/firebase/php-jwt.git",
+                "reference": "fa8a06e96526eb7c0eeaa47e4f39be59d21f16e1"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/firebase/php-jwt/zipball/fa8a06e96526eb7c0eeaa47e4f39be59d21f16e1",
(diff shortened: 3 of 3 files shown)


bshaffer:
@AwesomeBobX64 this is not the correct way to handle this problem. `composer.lock` is *not* supposed to be committed to utility libraries such as these, and only to deployed applications.

What you need to do is declare a `composer.json` in your own application, which will generate its own `composer.lock` that you can then commit. Because this library locks to `@dev`, it's actually allowing you to define `v0.2` OR `v0.3` in your own project. So in your project, your `composer.json` should look something like this:

```json
{
    "require": [
        "google/apiclient":"^2.0",
        "google/apiclient-services":"v0.2"
    ]
}
```

AlecHFerguson:
I'm confused as to why this PR was declined.  I ran into a breaking API change between minor versions of your package; a bug that can be prevented by the exact changes proposed here.

Bundler, npm, and most other dependency managers recommend committing the lockfile, and Composer is no exception:
https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

The justification is simple: 3rd party packages change, so don't let those breaking changes creep into your project.  The solution you're proposing requires that users of the Google API Client have knowledge of your package's dependencies, a prospect that seems to undermine the case for using libraries.

bshaffer:
@AlecHFerguson If this library included a `composer.lock` file, it would not have any affect in your project. Instead, *your* library should include a `composer.lock` file to lock these dependencies.

The downstream `google/apiclient-services` repository has 100+ APIs which are often updated. They do not follow semantic versioning, which makes it impossible to provide both the flexibility to update to the latest version *and* the guarantee that updating won't break your code. 

The current implementation allows you to lock the downstream `google/apiclient-services` dependency to a version if you need to, or get the latest APIs when desired.

AwesomeBobX64:
That is not true. I don't believe you understand the semantic versioning specification that your package versions imply. Please check out semver.org and reconsider our pull requests.

bshaffer:
@AwesomeBobX64 
Your PR does two things:

1. Includes a `composer.lock` - this will have **no effect** for projects pulling this library in as a dependency. The linked-to article refers to committing the lock file for *applications*, but this repository is not an application, it is a dependency, and therefore does not need a lock file. 
2. Sets the dependency for `google/apiclient-services` to `v0.2`. This is too strict of a requirement, as the APIs are constantly updating.

AwesomeBobX64:
1. That composer.lock file WILL lock the consumers of this package to only download the dependency versions identified in the composer.lock file. That is the purpose of including the composer.lock file in version control.
2. If your dependencies aren't stable, lock down to stable versions via the composer.lock file. v0.3 vs v0.2 completely changed the interfaces of the classes and is a breaking change. This is why it is important to lock down these dependency versions.

An update in the auth package caused our dependency on the google/apiclient to fail because even though we had locked into 2.0.0 of the google/apiclient package, it's dependency on the google/auth package was loose enough that even though my code hadn't changed, the dependency code HAD and I was encountering bugs from the now updated google/auth package.

The dependencies _should_ be locked down in a composer.lock file because of problems like these. When you have tested new versions of your dependencies still work with your package, then release a new version of your package with an updated composer.lock file which locks in which dependencies were tested to work with the new version of your package.

This is not an uncommon practice with mature, versioned package development processes.

Please reconsider.

bshaffer:
I have explained the rationality behind these decisions, as well as how composer lock files work. There is nothing more to consider.

PR #1000, opened 2016-07-21
tag client services

bshaffer:
In preparation to BC changes about to be merged in https://github.com/google/google-api-php-client-services/pull/25, we will update this version of the library to depend on `google/apiclient-services:v0.5`


tmatsuo:
LGTM

PR #1019, opened 2016-08-03
updates auth and uses in-memory cache

bshaffer:


src/Google/AccessToken/Verify.php
```diff
@@ -55,8 +56,12 @@ public function __construct(ClientInterface $http = null, CacheItemPoolInterface
       $http = new Client();
     }
 
-    if (is_null($cache) && class_exists('Stash\Pool')) {
-      $cache = new Pool(new FileSystem);
+    if (is_null($cache)) {
+      if (class_exists('Stash\Pool')) {
```
tmatsuo: Can you create a utility for creating a cache and reuse it?

bshaffer: I'd really rather not. The logic is very minimal, and creating a utility would require a new class. Also, the logic here is only to preserve BC. Next time we break BC, I plan to make `MemoryCacheItemPool` the default, and remove the Stash logic all together.

tmatsuo: Hm, ok, I can live with that.

tmatsuo:
LGTM

Issue #1024, opened 2016-08-10
Unknown stream type php://stderr


eRadical:
Defaulting to Monolog is not ok on AppEngine

I get 
[message:protected] => The stream or file "php://stderr" could not be opened: failed to open stream: Unknown stream type php://stderr.

https://cloud.google.com/appengine/docs/php/logs/#PHP_writing_application_logs » says a simple call to syslog() would suffice!

@bshaffer - do you have any path on mitigating this? I might do it myself if I can have a hint/direction...

bshaffer:
Try this:

```php
$client = new Google_Client;
$logger = new Monolog\Logger('google-api-php-client');
$logger->pushHandler(new Monolog\Handler\SyslogHandler('app'));
$client->setLogger($logger);
```

The way to mitigate this would be to call `$this->isAppEngine()` in `Google_Client::createDefaultLogger`. If you want to submit a PR with tests for this, I'd be happy to merge it!

bshaffer:
I submitted #1025 to address this in the library. If you want to pull down my branch and let me know if it works for you, that would be greatly appreciated

```
composer require google/apiclient:dev-fix-default-logger-for-appengine
```

eRadical:
I'll test tomorrow to see if all is ok w/ Monolog\Handler\SyslogHandler and let you know!

eRadical:
All checks are ok!

PR #1025, opened 2016-08-10
uses syslog handler on appengine

bshaffer:
see #1024

### DIFF
--- a/tests/Google/ClientTest.php
+++ b/tests/Google/ClientTest.php
@@ -243,6 +243,26 @@ class Google_ClientTest extends BaseTest
     $this->assertInstanceOf('Google_Model', $dr_service->files->listFiles());
   }
 
+  public function testDefaultLogger()
+  {
+    $client = new Google_Client();
+    $logger = $client->getLogger();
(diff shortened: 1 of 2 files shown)


tests/Google/ClientTest.php
```diff
@@ -243,6 +243,26 @@ public function testPrepareService()
...
+
+  public function testDefaultLoggerAppEngine()
+  {
+    $_SERVER['SERVER_SOFTWARE'] = 'Google App Engine';
+    $client = new Google_Client();
+    $logger = $client->getLogger();
+    $this->assertInstanceOf('Monolog\Logger', $logger);
+    $handler = $logger->popHandler();
+    $this->assertInstanceOf('Monolog\Handler\SyslogHandler', $handler);
+    unset($_SERVER['SERVER_SOFTWARE']);
```
tmatsuo: Nit:
If the assert fails, will the `unset` statement still be executed?

tmatsuo:
LGTM with a minor nit

tests/Google/ClientTest.php
bshaffer: NOW it will! 😄

PR #1030, opened 2016-08-12
Correct DocBlock @throws to LogicException on Client.php

inverse:
Correct DocBlock @throws to LogicException


inverse:
I signed it!

bshaffer:
looks good, thanks for finding this!

Issue #1046, opened 2016-09-01
Uses of http_build_query should use the $arg_seperator parameter


chriscct7:
Certain hosting providers alter the default setting for PHP's `arg_separator.output` configuration to an HTML escaped value of `&` instead of the default `&` value. The result of this is that on these hosting providers, calls that utilize functions in both the legacy and new versions of the PHP client library, can fail. 

The quick and simple solution is that in the legacy and stable versions of the PHP client API, uses of `http_build_query` need to be updated to utilize the third parameter.

I'll submit a pr for the uses in each version shortly.

chriscct7:
Prs #1047 and #1048 opened. When both are merged, this issue is solved.

bshaffer:
Wow, I'm pretty shocked it's allowed to change the default behavior here. This would break many implementations, not just ours. Any idea what the use case is?  I assume this was added for sites using PHP as a glorified templating language.

chriscct7:
> Wow, I'm pretty shocked it's allowed to change the default behavior here. 

Yeah, I was as well. We use the PHP Google API for a our Google Analytics plugin that we write for the WordPress platform, and we probably get a dozen tickets a week at least from this issue (the currently deployed version uses an older version of the API which had more of these calls in it; we're getting ready to launch a new version that uses the latest & greatest). From what we can tell, from talking to a couple of the hosting companies involved like HostEurope, they do it as "a security measure" because substituting an escaped value of ampersand somehow makes it safer. I presume its assuming a usecase of people outputting it as a template language as well, but I don't fully understand their usecase either (I just know it completely breaks things). Who knows. I know WordPress has hardcoded the third parameter of all of the core http_build_query calls to the correct default value for this reason.

chriscct7:
I should note, we were originally just going to wrap an ini_set to force it back whenever our plugin was loaded, but give there's only 1 call to it in the new API, we've gone ahead and done these prs on our fork of the PHP API which we're shipping with our rewrite.

chriscct7:
I should also note that it's not just hosts doing this. There are older versions of Drupal where the Drupal application actually does this: https://www.drupal.org/node/2476237

bshaffer:
I agree this is a better solution than calling `ini_set`. ini parameters should never be set by a 3rd party.

PR #1047, opened 2016-09-01
Update the Google AccessToken Revoke class to use the `arg_seperator` param in the http_build_query call

chriscct7:
Solves https://github.com/google/google-api-php-client/issues/1046 for the new API version

### DIFF
--- a/src/Google/AccessToken/Revoke.php
+++ b/src/Google/AccessToken/Revoke.php
@@ -58,7 +58,7 @@ class Google_AccessToken_Revoke
       }
     }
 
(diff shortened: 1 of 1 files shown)


chriscct7:
I signed it!

bshaffer:
I'm a little confused as to what effect this has, since we provide a single parameter, and so will not require a separator. i.e. this will always output the following, regardless of what the default separator is set to:

```
string() "token=TOKENSTRING"
```

It would be great to have a test to cover this, so we can be sure it'll fix the issue.

chriscct7:
Indeed you are correct on this. Didn't notice the v2 API always only has 1 parameter. I think you can close this.

PR #1048, opened 2016-09-01
Update the Google HTTP Request class to use the `arg_seperator` in the http_build_query call

chriscct7:
Solves https://github.com/google/google-api-php-client/issues/1046 for the legacy version

### DIFF
--- a/src/Google/Http/Request.php
+++ b/src/Google/Http/Request.php
@@ -416,7 +416,7 @@ class Google_Http_Request
   {
     $str = '';
     $path = parse_url($this->getUrl(), PHP_URL_PATH) . "?" .
-        http_build_query($this->queryParams,'', '&');
+        http_build_query($this->queryParams, '', '&');
     $str .= $this->getRequestMethod() . ' ' . $path . " HTTP/1.1\n";
 
     foreach ($this->getRequestHeaders() as $key => $val) {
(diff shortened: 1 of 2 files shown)


chriscct7:
I signed it!

owneroverall:
Hello geller feller

Den 17:31 tor 1 sep 2016googlebot <notifications@github.com> skrev:

bshaffer:
It would be great to have test coverage for this.

chriscct7:
So I'm thinking for a test, as opposed to testing this exact case, we could ini_set the value to `&amp` in like the bootstrap, because hypothetically, if everything is working correctly nothing should fail if that is done. Or did you have a specific approach in mind?

bshaffer:
I'd rather do this:

```php
// tests/Google/AccessToken/VerifyTest.php
//...
public function testHttpBuildQuery()
{
    $argSeparatorOutput = ini_get('arg_separator.output');
    ini_set('arg_separator.output', '&amp;');
    // do test 
    ini_set('arg_separator.output', $argSeparatorOutput);
}
//...
```

bshaffer:
@chriscct7 alternatively, you can give me [collaboration access to your fork](https://github.com/blog/2247-improving-collaboration-with-forks) and I'll make the change.

chriscct7:
@bshaffer sorry for the delay. I've enabled the edit from maintainers and have added unit tests for both the normal and `&amp;` values of the arg_seperator. Unit test and fixes have been verified against an unfixed fork of repo that was update with tests.

bshaffer:
Sorry for the delay as well! Merged, thank you.

chriscct7:
Not a problem!

PR #1062, opened 2016-09-27
Added missing q parameter to filter the drafts list (users_drafts)

### LINKED ISSUE #1061: Cannot query the drafts list using listUsersDrafts() method
The Gmail API supports an optional parameter 'q' to filter the list of drafts. 

https://developers.google.com/gmail/api/v1/reference/users/drafts/list

But the client library doesn't support this optional 'q' parameter. When I pass this parameter, it gives the error, An error occurred: (list) unknown parameter: 'q'. It doesn't let me query the drafts for a given keyword and filter the results.

Nimeshka:
this commit fixes #1061

### DIFF
--- a/src/Google/Service/Gmail.php
+++ b/src/Google/Service/Gmail.php
@@ -179,6 +179,10 @@ class Google_Service_Gmail extends Google_Service
                   'location' => 'query',
                   'type' => 'string',
                 ),
+                'q' => array(
+                  'location' => 'query',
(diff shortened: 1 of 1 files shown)


bshaffer:
This file is auto-generated, and so cannot be modified with a pull request. I'll run the code generator and see if this addresses your issue.

Also please upgrade to the `master` branch, as `v1-master` is deprecated.

PR #1063, opened 2016-09-27
update firebase/php-jwt to 4.0

ashbeats:


ashbeats:
ok. it's signed.

bshaffer:
Is there a specific reason you need Firebase `4.0`?

boredom2:
I need it:-)
The Microsoft Azure SDK has been updated and is only compatible with JWT 4.0 now.
So its really complicated to include Google and Azure with different dependencies for this Lib.

ashbeats:
I'm using laravel/spark, which requires firebase v4.0.

tomschlick:
Another request for this from another laravel/spark user. Can't upgrade spark without it.

Doesn't look like there are any breaking changes in the `3.0`->`4.0` migration so are there any reasons not to upgrade?

etremblay:
This PR would be very appreciated !

bshaffer:
This change will not have an affect as long as `google/auth` requires `~3.0`. This change should fix that: https://github.com/google/google-auth-library-php/pull/137. I will merge this as soon as this is tagged and updated

tomschlick:
Any update? google/google-auth-library-php#137 has been merged.

bshaffer:
This was merged into master in #1082

PR #1066, opened 2016-10-04
Fix the approval_prompt default value documentation.

nathanmarcos:
Fixed "setApprovalPrompt" method documentation to the code right default.

Here you can see the default value to approval_prompt is "auto": [Client.php#L129](https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L129)


nathanmarcos:
I signed it!

bshaffer:
Thanks!

PR #1069, opened 2016-10-07
Don't distribute tests, build files etc. for composer

willemstuursma:
Don't distribute tests, build files etc. when installing with Composer with `--prefer-dist` (the default). This saves lots of bandwith and file extraction when installing dependencies with default settings.


bshaffer:
@willemstuursma thank you!

PR #1078, opened 2016-10-14
fix for the proxy headers quirks check

spescina:
The check for applying the quirk workaround was switched


spescina:
I signed it!

bshaffer:
The comment above the code block essentially says "Subtract the proxy header size for cURL `>=7.30.0`". `needsQuirk` returns true when the cURL version is `<7.30.0`. 

```php
        // Subtract the proxy header size unless the cURL bug prior to 7.30.0
        // is present which prevented the proxy header size from being taken into
        // account.
        if (!$this->needsQuirk()) {
            $headerSize -= strlen($established_header);
        }
```
So the current code is accurate.

Issue #1075, opened 2016-10-12
Sporadic "Invalid Credentials" Issues - Service Account


omnicodagithub:
I seem to be having weird issues with "Invalid Credentials" errors, service accounts and the Search Console API.  Sometimes everything works and I can access the API, then some time later I may try an access the SC API and I get an "Invalid Credentials" error. This is without changing any code.

Edit: Very weird. I swapped around my scopes and then it started working again. Nothing else was changed. Why would swapping my scopes around every so often work?

Here is what I am using:

```
class Google {

    private function googleApiAuthorise()
    {

        $client = new \Google_Client();
        putenv('GOOGLE_APPLICATION_CREDENTIALS=' . base_path() . '/keys/Tool-xxxxxxxxx.json');
        $client->useApplicationDefaultCredentials();
        $client->setSubject(example@example.iam.gserviceaccount.com');
        $client->setApplicationName("Tool");
        $scopes = ['https://www.googleapis.com/auth/webmasters.readonly', 'https://www.googleapis.com/auth/webmasters'];
        $client->setScopes($scopes);

        if( $client->isAccessTokenExpired()) {

            $client->refreshTokenWithAssertion();

        }

        return $client;

    }

    public function getSearchAnalytics($search_type = 'web')
    {

        $authorise = Google::googleApiAuthorise();

        $service = new \Google_Service_Webmasters($authorise);

        $request = new \Google_Service_Webmasters_SearchAnalyticsQueryRequest;

        $request->setStartRow(0);
        $request->setStartDate('2016-06-01');
        $request->setEndDate(Carbon::now()->toDateString());
        $request->setSearchType($search_type);
        $request->setRowLimit(200);
        $request->setDimensions(array('date','page','country','device','query'));

        $query_search = $service->searchanalytics->query("http://www.example.com/", $request); 
        $rows = $query_search->getRows();

        return $rows;

    }
}
```

The weird thing is it works sometimes. Is this some sort of bug? I've followed pretty much every tutorial out there about service accounts and this API client, but it still seems to be quite random as to why sometimes I can use the API and other times I get an "Invalid Credentials" error.

dkrausahma:
When you say "sporadic" do you mean after 3600 seconds? That would be the time when the access token is expired and you would have to get a new one.

I am having the same issue and i do understand that it is due to the expired access token. This is my particular code sequence (which is quite similar to yours):

```php
[...]

PR #1085, opened 2016-10-20
removes stash logic, adds caching test for MemoryCache

bshaffer:


bshaffer:
Yet another way to solve #1075 

This is my preferred method. That way, users can handle `tedivm/stash` at their own risk.  However this leaves us w/o a recommended way use file caching.

tmatsuo:
sounds good too

mwithheld:
Docs on the GitHub main page (README.md) should be updated: remove the reference to Stash, replace with an example for MemoryCache.  Please include a link to the MemoryCache source.

bshaffer:
Would you mind filing an issue for this and we will get the readme updated?

PR #1091, opened 2016-10-28
Fix unused params

danielemontesi:
Fix unused params in a test class


bshaffer:
Good catch! Thank you.

PR #1092, opened 2016-10-31
Fix some complexity in the call() method

borgogelli:
I have tried to remove some of the complexity of the call() method in the Resource class

### DIFF
--- a/src/Google/Service/Resource.php
+++ b/src/Google/Service/Resource.php
@@ -80,15 +80,7 @@ class Google_Service_Resource
   public function call($name, $arguments, $expectedClass = null)
   {
     if (! isset($this->methods[$name])) {
-      $this->client->getLogger()->error(
-          'Service method unknown',
-          array(
-              'service' => $this->serviceName,
-              'resource' => $this->resourceName,
-              'method' => $name
-          )
-      );
-
+      $this->logServiceError('Service method unknown', $name);
       throw new Google_Exception(
           "Unknown function: " .
           "{$this->serviceName}->{$this->resourceName}->{$name}()"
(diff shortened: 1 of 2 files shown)


borgogelli:
I signed it

borgogelli:
I signed it

src/Google/Service/Resource.php
```diff
@@ -232,6 +200,37 @@ public function call($name, $arguments, $expectedClass = null)
...
 
+  private function logServiceError($txt, $name, $parameter = null)
+  {
+    $array = array(
+        'service' => $this->serviceName,
+        'resource' => $this->resourceName,
+        'method' => $name,
+        'parameter' => $parameter
+    );
+    if ($parameter) {
```
bshaffer: this conditional is not necessary, please remove it!

src/Google/Service/Resource.php
```diff
@@ -232,6 +200,37 @@ public function call($name, $arguments, $expectedClass = null)
...
+  }
+  
+  private function logServiceInfo($txt, $name, $parameters)
+  {
+    $array  = array(
+            'service' => $this->serviceName,
+            'resource' => $this->resourceName,
+            'method' => $name,
+            'arguments' => $parameters,
+        );
```
bshaffer: This indentation is off - this should be back four spaces, and the array arguments should only be indented 2 spaces.

src/Google/Service/Resource.php
```diff
@@ -232,6 +200,37 @@ public function call($name, $arguments, $expectedClass = null)
...
+    }
+    $this->client->getLogger()->error(
+        $txt,
+        $array
+    );
+  }
+  
+  private function logServiceInfo($txt, $name, $parameters)
+  {
+    $array  = array(
```
bshaffer: Let's use a variable name like `$context`, to match with the [PSR spec](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#3-psrlogloggerinterface).

src/Google/Service/Resource.php
```diff
@@ -232,6 +200,37 @@ public function call($name, $arguments, $expectedClass = null)
     return $this->client->execute($request, $expectedClass);
   }
 
+  private function logServiceError($txt, $name, $parameter = null)
```
bshaffer: Let's make the method signature look like this:

```
private function log($level, $txt, $name, $parameter = null)
```

Then we can make "info" and "error" the first argument to these calls, and consolidate two methods into one.

bshaffer:
As I have not heard back, closing this PR.

Issue #1064, opened 2016-09-27
Refresh token doesn't get returned in $client->fetchAccessTokenWithRefreshToken($refreshToken)


rolanyang:
While running this example

https://developers.google.com/sheets/quickstart/php

I noticed after the AccessToken expires it uses refresh_token to fetch a new token however it returns creds without the refresh_token included. As a result, subsequent attempts to fetch a new accessToken (after the expire) fail because the refresh_token is not saved and thus not able to be passed back to fetchAccessTokenWithRefreshToken again.

Should this be reinserted back into the creds?

```
/vendor/google/apiclient/src/Google/Client.php   2016-09-27 15:13:43.178769896 -0700
@@ -271,6 +271,7 @@
     $creds = $auth->fetchAuthToken($httpHandler);
     if ($creds && isset($creds['access_token'])) {
       $creds['created'] = time();
+      $creds['refresh_token'] = $refreshToken;
       $this->setAccessToken($creds);
     }

```

JsChiSurf:
Just ended up here as a result of a search after finding the same thing.

LetterboxDelivery:
Also seeing the same problem

bshaffer:
@LetterboxDelivery is right - the refresh token is only delivered the first time an application is authorized. 

We could add the refresh token back to the in-memory access token, but this is not a guaranteed fix, as the access token can be [cached at the auth level](https://github.com/google/google-auth-library-php/blob/master/src/FetchAuthTokenCache.php). I do think it would be an improvement, however.

JsChiSurf:
The important point for me is, I had setup access to the Calendar API back in January, and downloaded this library, and that version of the API, when the access token has expired, renews and passes back both the new token and the (original) refresh token, for saving / caching the file off for use after the next expiration, using:

// Refresh the token if it's expired.
if ($client->isAccessTokenExpired()) {
  $client->refreshToken($client->getRefreshToken());
  file_put_contents($credentialsPath, $client->getAccessToken());
}

Has this behavior intentionally changed in this version?  The current version calls (by way of example in the quick start):

// Refresh the token if it's expired.
if ($client->isAccessTokenExpired()) {
  $client->fetchAccessTokenWithRefreshToken($client->getRefreshToken());
  file_put_contents($credentialsPath, json_encode($client->getAccessToken()));
}

[...]

PR #1121, opened 2017-01-04
ensures refresh token is added back to the access token, adds tests

bshaffer:
addresses issue #1064

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -272,6 +272,9 @@ class Google_Client
     $creds = $auth->fetchAuthToken($httpHandler);
     if ($creds && isset($creds['access_token'])) {
       $creds['created'] = time();
+      if (!isset($creds['refresh_token'])) {
+        $creds['refresh_token'] = $refreshToken;
+      }
       $this->setAccessToken($creds);
     }
 
--- a/tests/Google/ClientTest.php
+++ b/tests/Google/ClientTest.php
@@ -462,7 +462,82 @@ class Google_ClientTest extends BaseTest
     $client->setHttpClient($http);
     $client->fetchAccessTokenWithRefreshToken("REFRESH_TOKEN");
     $token = $client->getAccessToken();
-    $this->assertEquals($token['id_token'], "ID_TOKEN");
+    $this->assertEquals("ID_TOKEN", $token['id_token']);
+  }
+
+  /**
+   * Test that the Refresh Token is set when refreshed.
+   */
+  public function testRefreshTokenIsSetOnRefresh()
+  {
+    $refreshToken = 'REFRESH_TOKEN';
+    $token = json_encode(array(
+        'access_token' => 'xyz',
+        'id_token' => 'ID_TOKEN',
+    ));
+    $postBody = $this->getMock('Psr\Http\Message\StreamInterface');
+    $postBody->expects($this->once())
+      ->method('__toString')
+      ->will($this->returnValue($token));
+    $response = $this->getMock('Psr\Http\Message\ResponseInterface');
+    $response->expects($this->once())
+      ->method('getBody')
+      ->will($this->returnValue($postBody));
+    $http = $this->getMock('GuzzleHttp\ClientInterface');
+    $http->expects($this->once())
+      ->method('send')
+      ->will($this->returnValue($response));
+
+    if ($this->isGuzzle5()) {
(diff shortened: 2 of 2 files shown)


tomaspinho:
looks great! thanks for the response ;)

jonnywilliamson:
Sorry a little confused here.

If I have a `refresh_token` and pass it to the client along with the `access_token`, and the client detects that the `access_token` is expired and magically/automatically gets me a NEW one using the `refresh_token`...

Why would you need to get the `refresh_token` returned to you again? Surely it doesn't expire (unless revoked?)

Did I misunderstand that PR #1121 ?

stevenscg:
Many refresh tokens do expire, though typically after a large amount of time (weeks, months, years, etc). Also for some systems, the refresh token may be expired after it has been used once. I can't remember if that is per the spec or not.

I ran into the lack of refresh token issue immediately when I started using this library to access the gmail API in "offline" mode (or whatever it is called).

jonnywilliamson:
OK,

I must have been confused when I saw this note at the bottom on the google oauth playground:

image

jonnywilliamson:
>I ran into the lack of refresh token issue immediately when I started using this library to access the gmail API in "offline" mode (or whatever it is called).

I think a lot of people miss the fact that the `refresh_token` is only ever issued at the VERY start when the oauth has been complete for the first time.

It's like you're only chance to grab it and save it for later use. I think a lot of people believe it should be returned to them on every request, but that's not what google does....at least when I was making my app a year or so ago!

mikejichaow:
Can someone help verify when did this get fix? looks like the current commit is after v2.1.0. What's the plan to have another release? I'd love to have a new release asap since I don't have access to install composer yet. Thanks!

bshaffer:
@mikejichaow I have been waiting for confirmation that this fix worked before tagging a new release.

mikejichaow:
@bshaffer Hey, I thought you guys made tests already. Is there anyway I can get the composer processed autoload files so I can test it? Also, is there a way to set the parameter on how long does the token expire? I know on default its always 3600.

bshaffer:
@mikejichaow Install composer and do the following:

```
cd /path/to/my-project
composer install google/apiclient:dev-master
```

Then in your project you can include the autoloader:


```php
# /path/to/my-project/index.php
require_once __DIR__ . '/vendor/autoload.php';
```

mikejichaow:
@bshaffer I've been trying to get composer working but got the following errors:
` Problem 1
    - The requested package google/apiclient No version set (parsed as 1.0.0) is satisfiable by google/apiclient[No version set (parsed as 1.0.0)] but these conflict with your requirements or minimum-stability.`
  `Problem 2
    - The requested package minimum-stability could not be found in any version, there may be a typo in the package name.`
What are the settings I should use for composer.json?

bshaffer:
Try setting the version as `^2.0`: 

```sh
composer require google/apiclient:^2.0
```

bshaffer:
@mikejichaow you can also now download the release manually [here](https://github.com/google/google-api-php-client/releases/tag/v2.1.1)

mikejichaow:
Got it. Thank you. 
To test it though, is there a way to set expiration time to be shorter/longer? I tried to just call fetchAccessTokenWithRefreshToken before it expires. It passed and logged in but did not seem to update the token. Even though the file modified time is changed which means it got written to. The token is still the same. Will it only got updated after 3600 s?

bshaffer:
@mikejichaow there is no way to change the access token expiration, as it comes from the Google oauth server. You could try hacking the cache file and manually changing the expire time, but I'd say just wait 3600 seconds :=P

Micle86:
bshaffer я считаю что файл должен содержать refresh token

bshaffer:
@Micle86 Я считаю, что это делает !

aaronhatton:
I have updated using Composer and have also downloaded the release manually and am still having issues, please can you advise. Error message I am getting daily is:

`Fatal error: Uncaught LogicException: refresh token must be passed in or set as part of setAccessToken`

mikejichaow:
@arronhatton did u check if you have the refresh token for the first time u make a call? and the second time (after expired) you should call again and it will give u back the refresh token. my test using it works fine now.

aaronhatton:
Found my issue, was using OAuth 2.0 for web server instead of server to server

PR #1122, opened 2017-01-04
ensures JWT leeway is restored after we set it

bshaffer:


bshaffer:
addresses concerns in https://github.com/google/google-api-php-client/pull/828#issuecomment-270470977 regarding `JWT::$leeway`'s global scope. cc @martinfree

calebdelnay:
There is a bug in this PR. The code adds a `'jwt'` config key but then uses `$this->jwt` instead of the correct `$this->config['jwt']`.

bshaffer:
@calebdelnay thank you for catching this!

bshaffer:
fixed in 8e3ac1a

bshaffer:
fixed in 8e3ac1a

PR #1128, opened 2017-01-11
update google/apiclient-servies to v0.9

bshaffer:


shalomabitan:
Hi,

Can you bump google-api-services to v0.11 I'm having some issues with the code base the would be fixed by v0.11.

 For example, adding parameter field `historyTypes` to `$service->users_history->listUsersHistory($userId, $opt_param);` throws an exception `(list) unknown parameter: 'historyTypes'` which is fixed by version 0.11

Thanks :)

PR #1137, opened 2017-01-19
fix createBatch call

bshaffer:


stevenscg:
@bshaffer Was `createBatch` working before this PR? I'm on `v2.1.2` and about to build some batch operations. Wondering if I need to run against master or wait for a point release or can stay on 2.1.2 for now.

bshaffer:
@stevenscg createBatch was working for any APIs with the batch path `/batch`. However, there are a handful of APIs that have a distinct "batchPath" declared in their discovery JSON. For those APIs, this change is required. To verify if the API was affected, check the discovery JSON, and if the property `batchPath` exists and is set to something other than `/batch`, you will need this version.

stevenscg:
Will do. Haven't had any issues with the Gmail API so far, but will upgrade when the release is cut. Thanks!

PR #1148, opened 2017-02-08
Update REST.php

mavelar:
Validate response code if is >= 500, I suggest to handle in our side if a 4xx error code or less happens.

### DIFF
--- a/src/Google/Http/REST.php
+++ b/src/Google/Http/REST.php
@@ -110,7 +110,7 @@ class Google_Http_REST
     $code = $response->getStatusCode();
 
     // retry strategy
-    if ((intVal($code)) >= 400) {
+    if ((intVal($code)) >= 500) {
       // if we errored out, it should be safe to grab the response body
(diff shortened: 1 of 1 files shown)


mavelar:
I signed it!

bshaffer:
I'm not sure why the original author of this code wanted to retry `400`-level errors. I cannot think of a scenario where this would be appropriate.

bshaffer:
I think this is unintended behavior. We do want to throw an exception for `400`-level errors, so the fix above is not quite what we're looking for.

tmatsuo:
Re: retry on 40x

Some cases (I think only) for 403, it's recommended.

Example:
https://developers.google.com/analytics/devguides/reporting/core/v3/errors

> 403	userRateLimitExceeded: Retry using exponential back-off. You need to slow down the rate at which you are sending the requests.

bshaffer:
This would be more exponential backoff than retry though, and is a special case.

PR #1174, opened 2017-03-15
[Minor] Static Code Analysis with Php Inspections (EA Extended)

kalessil:
I recently run into the "only Php talk" video and thought that supporting our community member would be a good idea =)

Tweaks found and fixed automatically by [Php Inspections (EA Extended)](https://github.com/kalessil/phpinspectionsea) (my pet-project):

- dead code;
- control flow simplifications;
- language level and code style;

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -53,11 +53,11 @@ class Google_AccessToken_Verify
       CacheItemPoolInterface $cache = null,
       $jwt = null
   ) {
-    if (is_null($http)) {
+    if (null === $http) {
       $http = new Client();
     }
 
-    if (is_null($cache)) {
+    if (null === $cache) {
       $cache = new MemoryCacheItemPool;
     }
 
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -477,10 +477,7 @@ class Google_Client
     }
 
     // If the token is set to expire in the next 30 seconds.
-    $expired = ($created
-      + ($this->token['expires_in'] - 30)) < time();
-
-    return $expired;
+    return ($created + ($this->token['expires_in'] - 30)) < time();
   }
 
   public function getAuth()
@@ -764,8 +761,8 @@ class Google_Client
     if (empty($this->requestedScopes)) {
       return null;
     }
-    $scopes = implode(' ', $this->requestedScopes);
-    return $scopes;
+
+    return implode(' ', $this->requestedScopes);
   }
 
   /**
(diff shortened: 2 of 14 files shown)


kalessil:
I signed it!

src/Google/AccessToken/Verify.php
```diff
@@ -53,11 +53,11 @@ public function __construct(
       CacheItemPoolInterface $cache = null,
       $jwt = null
   ) {
-    if (is_null($http)) {
+    if (null === $http) {
       $http = new Client();
     }
 
-    if (is_null($cache)) {
+    if (null === $cache) {
```
bshaffer: Why is this preferable to `is_null`?

src/Google/Client.php
```diff
@@ -366,7 +366,7 @@ public function authorize(ClientInterface $http = null)
     } elseif ($token = $this->getAccessToken()) {
       $scopes = $this->prepareScopes();
       // add refresh subscriber to request a new token
-      if ($this->isAccessTokenExpired() && isset($token['refresh_token'])) {
+      if (isset($token['refresh_token']) && $this->isAccessTokenExpired()) {
```
bshaffer: Why is this order preferable?

src/Google/Task/Runner.php
```diff
@@ -269,8 +264,7 @@ public function allowedRetries($code, $errors = array())
       return $this->retryMap[$code];
     }
 
-    if (!empty($errors) && isset($errors[0]['reason']) &&
-        isset($this->retryMap[$errors[0]['reason']])) {
+    if (!empty($errors) && isset($errors[0]['reason'], $this->retryMap[$errors[0]['reason']])) {
```
bshaffer: This should be under 80 characters, so it can be wrapped like this:


```php
    if (
        !empty($errors) && 
        isset($errors[0]['reason'], $this->retryMap[$errors[0]['reason']])
    ) {
```

src/Google/AccessToken/Verify.php
kalessil: `null === $cache` does the same, but without operating on calls stack, so it simply [faster](https://hakre.wordpress.com/2011/03/29/php-is_null-vs-null/).
And de-facto using is_null already judged as a bad practice.

src/Google/Client.php
kalessil: `isset()` faster, e.g. Symfony used it for some performance optimization, so we just put "faster" to fail part of condition at first place. Micro-optimization I would say.

src/Google/Task/Runner.php
kalessil: Oh, good catch - fixing it.

PR #1183, opened 2017-03-21
Require composer 6.x instead of 6.0.x

### LINKED ISSUE #1182: Support Guzzle 6.x
Guzzle 6.2.1 was released almost a year ago and it addresses some important [issues](https://github.com/guzzle/guzzle/releases/tag/6.2.1). However, this library only supports 5.2x and 6.0.x. I believe there's no breaking changes in the 6.x versions, so maybe a change to `^6.0` in the `composer.json` file should be enough.

dluces:
Fixes #1182

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "firebase/php-jwt": "~2.0|~3.0|~4.0",
         "monolog/monolog": "^1.17",
         "phpseclib/phpseclib": "~0.3.10|~2.0",
-        "guzzlehttp/guzzle": "~5.2|~6.0",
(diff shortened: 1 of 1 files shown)


dluces:
I signed it!

bshaffer:
In this context, the tilde (`~`) and caret (`^`) mean the same thing. They only diverge with versions `< 1.0` (see [docs](https://getcomposer.org/doc/articles/versions.md#tilde))

If you install and inspect the `composer.lock` file, you'll see the following:

```
"name": "guzzlehttp/guzzle",
 "version": "6.2.3",
```

So the current version supports Guzzle 6.2.

PR #1197, opened 2017-04-10
Fix two Cross-Site scripting vulnerabilities (XSS)

Neven-Biruski-DefenseCode:
Sample URLs:
http://vulnerablesite.com/google-api-php-client/examples/url-shortener.php/%22%3E%3Cscript%3Ealert('xss')%3C/script%3E%3Cdc
http://vulnerablesite.com/google-api-php-client/examples/url-shortener.php/%22%3E%3C/script%3E%3Cscript%3Ealert(42)%3C/script%3E?url=http%3A%2F%2Fwww.defensecode.com

During the security analysis, DefenseCode ThunderScan SAST discovered Cross Site Scripting vulnerabilities in Google's PHP client library for accessing Google APIs (google-api-php-client). The vulnerabilities were found in the sample code for using the Google's URL Shortener. 

Due to missing nonce token, the vulnerability is also directly exposed to Cross site request forgery (CSRF) attacks.

Vulnerability was found by Leon Juranic using DefenseCode ThunderScan (http://www.defensecode.com).

### DIFF
--- a/examples/url-shortener.php
+++ b/examples/url-shortener.php
@@ -115,7 +115,7 @@ if ($client->getAccessToken() && isset($_GET['url'])) {
     <a class='login' href='<?= $authUrl ?>'>Connect Me!</a>
   </div>
 <?php elseif (empty($short)): ?>
-  <form id="url" method="GET" action="<?= $_SERVER['PHP_SELF'] ?>">
+  <form id="url" method="GET" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
     <input name="url" class="url" type="text">
     <input type="submit" value="Shorten">
   </form>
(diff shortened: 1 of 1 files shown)


Neven-Biruski-DefenseCode:
I signed it!

bshaffer:
The bot seems to think you did not sign the CLA. Also, these examples are not really vulnerable to XSS, they are only run locally. Also I am unsure how using `htmlspecialchars` around `PHP_SELF` has anything to do with XSS.

Neven-Biruski-DefenseCode:
The easiest way to confirm the vulnerabilities is to visit the provided sample URLs. Even if you only run them on loopback adapter (which usually makes little sense for the production environment ;), the vulnerability can still be abused from any site on the Internet due to missing CSRF protection.

htmlspecialchars (according to the php manual) converts special characters to HTML entities. For example, if you take a look at the sample URLs in my first comment (and URL decode them, as php does), by using htmlspecialchars(), you can ensure that during the creation of the response certain characters from the URL become escaped - " would become &quot;, < &lt; and > &gt;. That will stop most XSS attacks. 

If the PHP_SELF itself was the point of misunderstanding, please check out how certain webservers handle extra characters after the name of the php script. For example, Apache will happily accept url such as vulnerablesite.com/script.php/"></script><script>alert(42)</script> , execute script.php, and the javascript from the url will still end up in PHP_SELF variable.

Feel free to ask for more clarification if needed.

sirdarckcat:
@Neven-Biruski-DefenseCode  thank you for your report. We are tracking this bug internally at b/37360079.

PR #1202, opened 2017-04-18
Update composer require ^2.0 to ^2.1

komirad:
Instructions to install ^2.0 is incompatible with other libraries due to google/auth version conflict. `Can only install one of: google/auth[v0.8, v0.11.1].`


bshaffer:
This change actually makes the library requirements *more* strict and not less, so this would not resolve a version conflict.

komirad:
Strangely, after I made the change I managed to install without version
conflict.

PR #1209, opened 2017-05-02
Added check for if range exists when getting progress for resumable u…

ytn3rd:
As per the [resumable upload docs](https://developers.google.com/youtube/v3/guides/using_resumable_upload_protocol#Handle_Response_to_Status_Request)

> If nothing has been uploaded yet, the API response will not include the Range header.

If no data has been uploaded you will get an invalid offset 1 when trying to access $range[1]. This fix allows progress to be set only if the range exists in the response header.


ytn3rd:
I signed it!

ytn3rd:
I signed it!

ytn3rd:
I signed it! (third times a charm right?)

PR #1212, opened 2017-05-02
WIP: Update url shortener sample

michaelbausor:
cc @sirdarckcat

Addressing issues raised in https://github.com/google/google-api-php-client/pull/1197 (b/37360079)

### DIFF
--- a/examples/url-shortener.php
+++ b/examples/url-shortener.php
@@ -60,7 +60,13 @@ $service = new Google_Service_Urlshortener($client);
  * local access token in this case
  ************************************************/
 if (isset($_REQUEST['logout'])) {
-  unset($_SESSION['access_token']);
+  if (validateCsrfToken()) {
+    unset($_SESSION['access_token']);
+    unset($_SESSION['csrf_token']);
+  } else {
+    $invalidCsrf = true;
+    generateCsrfToken();
+  }
 }
 
 /************************************************
@@ -76,6 +82,9 @@ if (isset($_GET['code'])) {
 
   // store in the session also
   $_SESSION['access_token'] = $token;
+
+  // generate and store a new csrf token
+  generateCsrfToken();
 
   // redirect back to the example
   header('Location: ' . filter_var($redirect_uri, FILTER_SANITIZE_URL));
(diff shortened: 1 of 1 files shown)


examples/url-shortener.php
```diff
@@ -101,12 +110,30 @@
...
+    $_SESSION['access_token'] = $client->getAccessToken();
+  } else {
+    $invalidCsrf = true;
+    generateCsrfToken();
+  }
+}
+
+function generateCsrfToken()
+{
+  $_SESSION['csrf_token'] = bin2hex(openssl_random_pseudo_bytes(32));
```
sirdarckcat: you need to check the return value of the second parameter
http://php.net/manual/en/function.openssl-random-pseudo-bytes.php

michaelbausor: I am not sure what behaviour we want here in the case that the $crypto_strong parameter comes back false. Should we emit a warning? It seems wrong to fail completely, as that is not a great experience for users trying to test out the example code on their local machine.

examples/url-shortener.php
```diff
@@ -101,12 +110,30 @@
...
+}
+
+function validateCsrfToken()
```
bshaffer: This is great! However, we have a place for helper functions like this in [`examples/templates/base.php`](https://github.com/google/google-api-php-client/blob/master/examples/templates/base.php). Please move this function there.

examples/url-shortener.php
```diff
@@ -101,12 +110,30 @@
...
-  $_SESSION['access_token'] = $client->getAccessToken();
+if ($client->getAccessToken() && isset($_REQUEST['url'])) {
+  if (validateCsrfToken()) {
+    $url = new Google_Service_Urlshortener_Url();
+    $url->longUrl = $_REQUEST['url'];
+    $short = $service->url->insert($url);
+    $_SESSION['access_token'] = $client->getAccessToken();
+  } else {
+    $invalidCsrf = true;
+    generateCsrfToken();
```
bshaffer: We can just follow the model of the [other validation functions](https://github.com/michaelbausor/google-api-php-client/blob/5e085f1041467a5ffacf17663911bddceca51f22/examples/url-shortener.php#L26) if this fails:

```php
if (!validateCsrfToken()) {
  echo invalidCsrfTokenWarning();
  return;
}

// and in base.php
function invalidCsrfTokenWarning()
{
    return 'The CSRF token is invalid, your session probably expired. Please refresh the page ';
}
```

examples/url-shortener.php
```diff
@@ -115,18 +142,28 @@
...
 <?php else: ?>
   You created a short link! <br />
   <a href="<?= $short['id'] ?>"><?= $short['id'] ?></a>
   <div class="shortened">
     <pre><?php var_export($short) ?></pre>
   </div>
-  <a href="<?= $_SERVER['PHP_SELF'] ?>">Create another</a>
+  <a href="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">Create another</a>
+<?php endif ?>
+<?php if (isset($invalidCsrf) && $invalidCsrf): ?>
```
bshaffer: See above comment and we can get rid of this block.

examples/url-shortener.php
```diff
@@ -115,18 +142,28 @@
     <a class='login' href='<?= $authUrl ?>'>Connect Me!</a>
   </div>
 <?php elseif (empty($short)): ?>
-  <form id="url" method="GET" action="<?= $_SERVER['PHP_SELF'] ?>">
+  <form id="url" method="POST" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
     <input name="url" class="url" type="text">
+    <input type="hidden" name="csrf_token" value="<?php echo $_SESSION['csrf_token']; ?>" />
```
bshaffer: Let's make a new function `getCsrfToken` that generates this if it doesn't exist.

examples/url-shortener.php
```diff
@@ -60,7 +60,13 @@
  * local access token in this case
  ************************************************/
 if (isset($_REQUEST['logout'])) {
-  unset($_SESSION['access_token']);
+  if (validateCsrfToken()) {
```
bshaffer: is any of this necessary? I don't understand why we need to remove the csrf token on logout, or why the validate function needs to be called. At the very least we can just unset the token.

bshaffer:
closing in favor of #1240

PR #1240, opened 2017-06-12
Michaelbausor fix url shortener sample

bshaffer:


bshaffer:
addresses changes requested in #1212

PR #1259, opened 2017-07-07
Add notice about maintenance mode

danoscarmike:


danoscarmike:
@bshaffer @michaelbausor can you please review?

michaelbausor:
LGTM

chriscct7:
@danoscarmike + @bshaffer is this for the 5.2 compatible version only or for master as well? For those building new php based applications which client library hsould we be using?

danoscarmike:
@chriscct7 apologies for the slow response.  Rest assured we continue to support these libraries.  This notice means that the libraries in this repository are considered mature and complete; so we do not expect to add any more major features.  We will continue to support these libraries, fix bugs, and add minor features as required.  We will also continue to release new versions of all API libraries as they are available; and release new API libraries as new services become available.

For those building new php applications using Google Cloud APIs we also encourage you to look at the latest client libraries available on cloud.google.com and developed open-source here on GitHub at googlecloudplatform/google-cloud-php.

Thanks for reaching out!

PR #1261, opened 2017-07-07
removes repeated check when setting array params

bshaffer:


bshaffer:
resolved in 9bef2490ed545c9deac78b12efa38ec01527d5f3

PR #1265, opened 2017-07-14
hardcoded leeway is problematic

danielhiga:
https://github.com/google/google-api-php-client/issues/827

some servers require more than 1 second.
set in 1 second only if it's less 1 second.
JWT::$leeway can have a greater default value or can be set with greater value in other part of the application.


danielhiga:
I signed the CLA just before the pull request.

PR #1266, opened 2017-07-14
change value of leeway only if it is lower than 1

danielhiga:
Some servers need more than 1 second.
Change JWT::$leeway only if it has a value lower than 1.

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -207,9 +207,11 @@ class Google_AccessToken_Verify
       $jwtClass = 'Firebase\JWT\JWT';
     }
 
-    if (property_exists($jwtClass, 'leeway')) {
-      // adds 1 second to JWT leeway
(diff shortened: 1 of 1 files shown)


bshaffer:
If you do not want the `$leeway = 1` behavior, you can instantiate your own Jwt instance and pass it into the client constructor: 

```php
new Client_Service(null, null, new Firebase\JWT\JWT());
```

danielhiga:
https://github.com/firebase/php-jwt/blob/master/src/JWT.php
line 30:  public static $leeway = 0;
$leeway is not an instance variable, you change the value globally.

bshaffer:
Yes but passing in your own instance bypasses calling the `getJwtService` function.

danielhiga:
this work:

```
JWT:$leeway = LEEWWAY; //some value
$client = new Google_Client(['jwt' => new Firebase\JWT\JWT()]);
```

thank you

bshaffer:
Yes!! That's it. Thanks for the correction.

PR #1291, opened 2017-08-25
Fix directory and reports stop channel urls

ricardofontanelli:
Report stop channel and Directory stop channel methods uses a different service path then their default class. When the lib build the full service url, the service path is duplicated.


ricardofontanelli:
I signed it!

ricardofontanelli:
I signed it!

ricardofontanelli:
I signed it!

gnosis7bit:
Ricardo, check you git settings.

Example:
```
sudo git config --global user.email "your_email@gmail.com"

PR #1292, opened 2017-08-25
Fix directory and reports stop channel urls 

ricardofontanelli:
Report stop channel and Directory stop channel methods uses a different service path then their default class. When the lib build the full service url, the service path is duplicated.

### DIFF
--- a/tests/Google/Service/ResourceTest.php
+++ b/tests/Google/Service/ResourceTest.php
@@ -138,12 +138,35 @@ class Google_Service_ResourceTest extends BaseTest
     $this->assertEquals("https://sample.example.com/method/path", (string) $request->getUri());
     $this->assertEquals("POST", $request->getMethod());
   }
-
+  
+  /* Some Google Service (Google_Service_Directory_Resource_Channels and Google_Service_Reports_Resource_Channels) use a different servicePath value that should override 
+  the default servicePath value, it's represented by a / before the resource path. All other Services have no / before the path*/
+  public function testCreateRequestUriForASelfDefinedServicePath()
+  {
+    $this->service->servicePath = '/admin/directory/v1/';
+    $resource = new Google_Service_Resource(
+    $this->service,
+      'test',
+      'testResource',
+      array("methods" =>
+        array(
+          'testMethod' => array(
+            'parameters' => array(),
+            'path' => '/admin/directory_v1/watch/stop',
+            'httpMethod' => 'POST',
+          )
+        )
+      )
+    );
+    $request = $resource->call('testMethod', array(array()));
+    $this->assertEquals('https://test.example.com/admin/directory_v1/watch/stop', (string) $request->getUri());
+  }
+  
   public function testCreateRequestUri()
   {
-    $restPath = "/plus/{u}";
+    $restPath = "plus/{u}";
     $service = new Google_Service($this->client);
-    $service->servicePath = "http://localhost";
+    $service->servicePath = "http://localhost/";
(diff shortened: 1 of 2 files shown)


ricardofontanelli:
I signed it!

bshaffer:
How about a test for this?

ricardofontanelli:
@bshaffer, is it ok now? Please, where can I report that the service URL at these four pages should be fixed?
- https://developers.google.com/admin-sdk/directory/v1/reference/channels/stop: service URL and API Explorer;
- https://developers.google.com/admin-sdk/reports/v1/reference/channels/stop: service URL and API Explorer;
- https://developers.google.com/admin-sdk/directory/v1/guides/push#stopping-notifications: service URL;
-https://developers.google.com/admin-sdk/reports/v1/guides/push#stopping-notifications: service URL

tests/Google/Service/ResourceTest.php
```diff
@@ -138,12 +138,35 @@ public function testCallServiceDefinedRoot()
...
+    $this->assertEquals('https://test.example.com/admin/directory_v1/watch/stop', (string) $request->getUri());
+  }
+  
   public function testCreateRequestUri()
   {
-    $restPath = "/plus/{u}";
+    $restPath = "plus/{u}";
     $service = new Google_Service($this->client);
-    $service->servicePath = "http://localhost";
+    $service->servicePath = "http://localhost/";
```
bshaffer: these changes here make me very nervous to merge this change. This will fundamentally change how requests are made, and so is a BC break.

ricardofontanelli: Have a look ate the google/apiclient-services classes, all of that classes have the same structure in the constructor method, for example the Google_Service_Calendar: 
$this->rootUrl = 'https://www.googleapis.com/';
$this->servicePath = 'calendar/v3/';
$this->version = 'v3';
$this->serviceName = 'calendar';

It means that the rootUrl is the URI of the service (always ended with a slash) and in all service classes the servicePath property ends with "/", it means that the old test was testing a behavior that would never happen wioth the servicePath value! I've changed my local environment and there's no problem with the way that the requests are build (Drive. Calendar, Classroom and Directory)!

The the main porpose of this PR is that in every service class, each resource returns a property called path, and there are only two cases where the path property starts with "/", Google_Service_Directory > Google_Service_Directory_Resource_Channels > stop and Google_Service_Reports > Google_Service_Reports_Resource_Channels > stop. 

For these two cases, the servicePath property should be overrided for the path value to build the correct service, because the endpoint structure is different of all other class resources. With the current implementation, we can't stop channels signed to Directory and Reports services.

tests/Google/Service/ResourceTest.php
```diff
@@ -138,12 +138,35 @@ public function testCallServiceDefinedRoot()
     $this->assertEquals("https://sample.example.com/method/path", (string) $request->getUri());
     $this->assertEquals("POST", $request->getMethod());
   }
-
+  
+  /* Some Google Service (Google_Service_Directory_Resource_Channels and Google_Service_Reports_Resource_Channels) use a different servicePath value that should override 
+  the default servicePath value, it's represented by a / before the resource path. All other Services have no / before the path*/
```
bshaffer: Can we get this guy to follow coding standards? E.g. wrap at 80 characters and use proper comment blocks:

```
/**
 * Some Google Service (Google_Service_Directory_Resource_Channels and 
 * Google_Service_Reports_Resource_Channels) use a different servicePath value 
 * that should override the default servicePath value, it's represented by a / 
 * before the resource path. All other Services have no / before the path
 */

PR #1299, opened 2017-09-05
Correcting checking the expiration of the access token

andrey-helldar:
If you use the access token with the `offline` parameter, the timestamp is not returned by the server.

If you call the method `$client->isAccessTokenExpired()`, it always answers that the token expired.

Therefore, introduced changes that allow to correctly determine the "validity" of an unlimited access token.

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -456,6 +456,8 @@ class Google_Client
   {
     if (!$this->token) {
       return true;
+    } elseif (!isset($this->token['expires_in'])) {
+      return false;
     }
 
     $created = 0;
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -460,7 +460,9 @@ class Google_Client
   {
     if (!$this->token) {
       return true;
-    } elseif (!isset($this->token['expires_in'])) {
+    }
+    
+    if (!isset($this->token['expires_in'])) {
       return false;
     }
 


src/Google/Client.php
```diff
@@ -456,6 +456,8 @@ public function isAccessTokenExpired()
   {
     if (!$this->token) {
       return true;
+    } elseif (!isset($this->token['expires_in'])) {
```
bshaffer: elseif not required since the previous statement returns true. Just add another if.

bshaffer:
This logic operates under the assumption offline access returns non-expiring access tokens, which isn't the case. It should return a refresh token which can fetch additional access tokens.

andrey-helldar:
The problem is that using an offline token, nothing is returned.

bshaffer:
Is it possibly in a different property?

andrey-helldar:
I realized reading the table from API Google.Sheets using api key for web-servers.
I must say that before that I forgot to open access to this table. When access is opened - the data began to be read without problems.
BUT the problem was that when you called the method `isAccessTokenExpired()`, it always returned `true`.

bshaffer:
Yes, this is a problem. I'm not sure always returning "false" is the right way to fix it, however.

andrey-helldar:
Example:
There is a web application that can work with both client_id and api_key.
The choice of the authorization method is in the settings (the programmer himself sets the desired value).
Question: how does the code know about the expiration date of the key, if it does not know what it is using - client_id or api_key?

bshaffer:
I'm not sure. More investigation is required as to what the real issue is. The claim is that if `access_type=offline` is set in the token request, no `expires` comes back. This would be strange if true, as the only difference I'm aware of with offline access is a `refresh_token` comes back with the access token, which is a separate thing entirely.

We could try setting the `expires` parameter if it isn't set, and I believe google's default is 1 hour (again, we need to look into it). But really I don't have a firm grip on the real problem here.

dwsupplee:
@andrey-helldar sorry to see this open for so long. Is this still relevant for you? I'm going to close this out for now, but please let me know if you'd like us to take another look here.

andrey-helldar:
@dwsupplee yes, my PR solves the problem, but I haven't used the package for a long time.

PR #1317, opened 2017-10-09
Fix method documentation in Client.php

Phobetor:


Phobetor:
I signed it!

src/Google/Client.php
```diff
@@ -753,7 +763,7 @@ public function getScopes()
   }
 
   /**
-   * @return array
+   * @return string
```
bshaffer: good catch, should be `string|null` though

src/Google/Client.php
```diff
@@ -441,11 +440,16 @@ public function getAccessToken()
...
+  /**
+   * @return string|null
+   */
   public function getRefreshToken()
   {
     if (isset($this->token['refresh_token'])) {
       return $this->token['refresh_token'];
     }
+
+    return null;
```
bshaffer: What the point of this? By default methods return null if no value is returned

src/Google/Client.php
Phobetor: You are right. I fixed that.

src/Google/Client.php
Phobetor: This is about clean code mainly. An indication that the method returns null on purpose and not because someone forgot a statement. PhpStorm for example raises an issue when returns with and without a value are mixed.

Phobetor:
Hi @bshaffer
Thanks for reviewing the changes. I fixed the return type hint and explained the added return statement.

bshaffer:
Thank you @Phobetor !

PR #1322, opened 2017-10-24
fix certificate caching, fixes #1033

### LINKED ISSUE #1033: Certificate caching issue with 2.0.2?
I'm using 2.0.2 (installed via composer) and am finding that if you don't manually remove /tmp/stash each day, auth fails during verifyIdToken().  I'm not sure what's up - the cache items aren't due for expiration, so it's not just a failure to purge.  Anyone else seeing this?

redcapital:
Looking at http://www.php-fig.org/psr/psr-6/ it seems `getItem` accepts only one parameter. The expiration must be set on the `CacheItemInterface` instance

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -178,7 +178,7 @@ class Google_AccessToken_Verify
   {
     $certs = null;
     if ($cache = $this->getCache()) {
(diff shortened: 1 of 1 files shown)


redcapital:
I signed it!

bshaffer:
This looks great, thank you!

PR #1334, opened 2017-11-04
Updated to PHPUnit 5.3.*

gabriel-caruso:
I've update PHPUnit to `5.3.*`, and dropped support to PHP versions `5.4` and `5.5`, as [PHPUnit 5 requires PHP `5.6`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.0.md#removed-1).

### Why dropped support for PHP versions `5.4` and `5.5`?
These versions are [no longer supported by PHP](http://php.net/supported-versions.php).

### Why PHPUnit 5.3 and not higher?
Because in 5.4, [`getMock` method was deprecated](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.4.md#changed-2), and we will need to work on that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,18 +16,9 @@ cache:
     - $HOME/.composer/cache
 
 php:
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
-
-# Test lowest dependencies on PHP 5.4
-# (Guzzle 5.2, phpseclib 0.3)
-matrix:
-  include:
-    - php: 5.4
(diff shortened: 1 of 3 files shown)


bshaffer:
Thank you for your Pull Request! Unfortunately, even though PHP 5.4 and PHP 5.5 are no longer supported, we still have users on those versions. So unless there is a good reason to update it (a better reason than upgrading our PHPUnit version), we are going to continue supporting these versions of PHP.

PR #1354, opened 2017-12-05
Document that not all APIs support service accounts

deviantintegral:
This was a confusing onboarding experience with the YouTube API as following the SDK example for service accounts allows some, but not all [YouTube data API calls](https://developers.google.com/youtube/v3/docs/errors). Hopefully this clears things up as it sounds like other APIs including AdSense don't fully support service accounts either.


deviantintegral:
I signed it!

bshaffer:
This is a great idea, although if we do this we should list all the unsupported APIs

PR #1352, opened 2017-11-30
Refactoring tests

carusogabriel:
I've refactored some tests with [`assertCount`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertCount), [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), [`assertTrue`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertTrue) and [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull) methods.

PR #1355, opened 2017-12-06
Refactoring tests

carusogabriel:
Continuing to refactory tests, like #1352, I've used:
- `assertArrayHasKey` instead of `isset` function;
- `assertGreaterThan` for mathematical comparisions;
- `assertCount` instead of `count` function;
- `assertInternalType` avoiding using `is_*` functions.


bshaffer:
These are great! Thank you @carusogabriel !

PR #1361, opened 2017-12-27
Fix HTTPS Conditions

msolters:
Only bypass HTTPS when two strict conditions are met:

*  We are on a local dev environment
*  The request is destined for a pubsub API domain


msolters:
Oh, didn't mean top open this on the upstream branch! Disregard

PR #1376, opened 2018-01-24
Fix #1350 error with rsa crypt

### LINKED ISSUE #1350: phpseclib > 2.0.7
phpseclib update has changed from method loadKey, line 97 in 
 google-api-php-client/src/Google/AccessToken/Verify.php

If you composer update and pull in the latest phpseclib, your google-api-client library will not work

Saqibm128:
#1350 

Simple enough.

More details: apparently phpseclib changed the method from loadKey to load, so this change fixes this.

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -94,7 +94,7 @@ class Google_AccessToken_Verify
       $exponent = new $bigIntClass($this->jwt->urlsafeB64Decode($cert['e']), 256);
 
       $rsa = new $rsaClass();
-      $rsa->loadKey(array('n' => $modulus, 'e' => $exponent));
+      $rsa->load(array('n' => $modulus, 'e' => $exponent));
 
       try {
         $payload = $this->jwt->decode(


Saqibm128:
I signed it!

Saqibm128:
More details: apparently phpseclib changed the method from loadKey to load, so this change reflects that.

bshaffer:
Hello @Saqibm128 !
I am looking at the [phpseclib code](https://github.com/phpseclib/phpseclib/blob/2.0.9/phpseclib/Crypt/RSA.php#L1507) and I'm still seeing the function `loadKey`. Also, your [test is failing](https://travis-ci.org/google/google-api-php-client/jobs/333001482#L617) because it cannot find the function `load`.

Could you paste your code or a gist to a repro case?

Saqibm128:
Sorry I think I was using master instead of a specific version. On master, they made this change, which was causing this issue for me.

Saqibm128:
Specific line on master.
https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Crypt/RSA.php#L467

dwsupplee:
I agree with @bshaffer that all I see is the method name `loadKey` for the versions of phpseclib pinned by the client (`~0.3.10|~2.0`). 

@Saqibm128 which version of `google/apiclient` are you pinned to?

The following installs, but includes [a client back to before phpseclib was introduced into the library](https://github.com/google/google-api-php-client/tree/v1.1.8/src/Google), so I'm not sure where a conflict with `loadKey` would come up.

```json
"require": {
    "google/apiclient": "*",
    "phpseclib/phpseclib": "dev-master"
}
```

Pulling in a newer version conflicts, `google/apiclient v2.0.0 requires phpseclib/phpseclib ~2.0 -> satisfiable by phpseclib/phpseclib[2.0.x-dev].`

```json
"require": {
    "google/apiclient": "~2.0",
    "phpseclib/phpseclib": "dev-master"
}
```

I'm not sure there is anything we can do here.

Saqibm128:
Entirely my mistake. Accidentally manually placed wrong version of phpseclib into the vendor folder. Closing request. Thank you

PR #1405, opened 2018-03-27
Increase minimum PHP version to 5.6

### LINKED ISSUE #1398: Update minimum php version to (at least ?) 5.5.
Hello,
I am maintaining an old project (the kind of which you call "the monster" with a low voice) which runs on php 5.4 (I know, it's bad...) without composer.

The README says :

>## Requirements
>PHP 5.4.0 or higher

But it actually seems to require a php version higher than 5.5 as it relies on `GuzzleHttp` which calls the [`curl_reset()`](https://secure.php.net/manual/en/function.curl-reset.php) function which was added in 5.5.

This results in a Fatal error :
>Fatal error: Call to undefined function GuzzleHttp\Handler\curl_reset() in /path_to/google-api-php-client-2.2.1/vendor/guzzlehttp/guzzle/src/Handler/CurlFactory.php on line 78

Note that I didn't try to run it on a php 5.5 server, but maybe the actual required version is 5.6...

Cheers.

mattwhisenhunt:
Several vendor library requirements are now >=5.5.9 or ^5.5.9 and 5.5 is
unsupported by the core PHP team anyways.

Fixes #1398

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This client library is supported but in maintenance mode only.  We are fixing ne
 The Google API Client Library enables you to work with Google APIs such as Google+, Drive, or YouTube on your server.
 
 ## Requirements ##
-* [PHP 5.4.0 or higher](http://www.php.net/)
+* [PHP 5.6.0 or higher](http://www.php.net/)
 
 ## Google Cloud Platform APIs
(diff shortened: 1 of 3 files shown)


bshaffer:
Here is the state of PHP 5 usage overall. I can pull up internal numbers as well and send them to you personally. 

https://w3techs.com/technologies/details/pl-php/5/all

mattwhisenhunt:
Please do, that would be great, thank you.

PR #1436, opened 2018-06-06
Add 1 to leeway instead of setting it to 1.

### LINKED ISSUE #1172: Bug in code (leeway is set to 1 rather than being increased by 1)
In google-api-php-client/src/Google/AccessToken/Verify.php line 213 you are not adding 1 to JWT leeway like you mentioned in the comments but rather setting it to 1. Kindly fix it.

if (property_exists($jwtClass, 'leeway')) {
// adds 1 second to JWT leeway
$jwtClass::$leeway = 1;
}

### LINKED ISSUE #1430: Custom leeway value for JWT verification gets overwritten with '1'
This happens on the Verify.php file located [here](https://github.com/google/google-api-php-client/blob/master/src/Google/AccessToken/Verify.php) at the `getJwtService()` method

```
  private function getJwtService()
  {
    $jwtClass = 'JWT';
    if (class_exists('\Firebase\JWT\JWT')) {
      $jwtClass = 'Firebase\JWT\JWT';
    }

    if (property_exists($jwtClass, 'leeway')) {
      // adds 1 second to JWT leeway
      // @see https://github.com/google/google-api-php-client/issues/827
      $jwtClass::$leeway = 1;
    }
  }
```

The comment says that this was done to fix issue #827 and it says that it 'adds', but instead it overwrites. This causes lots of `BeforeValidException`s to be thrown because most servers require more than 1 sec of leeway.

For example, if I set the leeway like so `\Firebase\JWT\JWT::$leeway = 60;` and then call `verifyIdToken($id_token)`, my custom leeway will be overwritten with 1.

A simple fix would work like this:
`$jwtClass::$leeway += 1;`

This would respect custom values set by the user and have a default value of `1` for those that don't change the leeway.

Cheers.

mattwhisenhunt:
@JustinBeckwith
Fixes #1430
Fixes #1172

### DIFF
--- a/src/Google/AccessToken/Verify.php
+++ b/src/Google/AccessToken/Verify.php
@@ -214,7 +214,7 @@ class Google_AccessToken_Verify
     if (property_exists($jwtClass, 'leeway')) {
       // adds 1 second to JWT leeway
       // @see https://github.com/google/google-api-php-client/issues/827
-      $jwtClass::$leeway = 1;
+      $jwtClass::$leeway += 1;
     }
 
     return new $jwtClass;


JustinBeckwith:
Nice!  Can we add unit/regression test coverage here though?

scottybo:
@bshaffer please could you review and approve this fix - it's a massive roadblock for us

bshaffer:
I don't agree with this change. We should just add `if ($jwtClass::$leeway < 1)`, as the spirit of this change was to make sure there was at least a little leeway given.

PR #1438, opened 2018-06-09
Update index.php

Assimilationstheorie:
Remove Short-Opentags, because the option "--enable-short-tags" in php.ini are not unique configured...
Add simple HTML-Validation with "required"
Submit Button without value-att. are not cool!

### DIFF
--- a/examples/index.php
+++ b/examples/index.php
@@ -22,8 +22,8 @@
 <div class="api-key">
   <strong>You have not entered your API key</strong>
   <form method="post">
-    API Key:<input type="text" name="api_key" />
-    <input type="submit" />
+    API Key:<input type="text" name="api_key" required/>
+    <input type="submit" value="Submit"/>
   </form>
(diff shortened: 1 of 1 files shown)


Assimilationstheorie:
I signed it!

examples/index.php
```diff
@@ -41,4 +41,4 @@
   <li><a href="multi-api.php">An example of using multiple APIs.</a></li>
 </ul>
 
-<?= pageFooter(); ?>
+<?php echo pageFooter(); ?>
```
tmatsuo: How about the one for the pageHeader in the same file?

bshaffer:
As mentioned in [this comment](https://github.com/google/google-api-php-client/pull/1439#issuecomment-396994756), `<?=` is ok and we are keeping them.

I have never heard that a `submit` button without the value attribute is bad practice. Can you please point me to some resources saying so?

Adding the `required` tag is okay, but these samples are meant to be kept as simple as possible, so I am not really a fan. However, if you remove all the other stuff we can merge that.

PR #1439, opened 2018-06-09
Update url-shortner.php

Assimilationstheorie:
Who write this Code? No Exception Handling, no Validation.
Don't use short opener for examples. php.ini is not unique configured!
Add simple HTML-Required-Attr.

### DIFF
--- a/examples/url-shortener.php
+++ b/examples/url-shortener.php
@@ -112,33 +112,32 @@ if ($client->getAccessToken() && isset($_REQUEST['url'])) {
   $short = $service->url->insert($url);
   $_SESSION['access_token'] = $client->getAccessToken();
 }
-
 ?>
 
 <div class="box">
 <?php if (isset($authUrl)): ?>
   <div class="request">
-    <a class='login' href='<?= $authUrl ?>'>Connect Me!</a>
+    <a class='login' href='<?php echo $authUrl; ?>'>Connect Me!</a>
   </div>
 <?php elseif (empty($short)): ?>
   <form id="url" method="POST" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
-    <input name="url" class="url" type="text">
+    <input name="url" class="url" type="text" placeholder="URL" required>
     <input type="hidden" name="csrf_token" value="<?php echo getCsrfToken() ?>" />
     <input type="submit" value="Shorten">
   </form>
   <form id="logout" method="POST" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
-    <input type="hidden" name="logout" value="" />
+    <input type="hidden" name="logout" />
     <input type="hidden" name="csrf_token" value="<?php echo $_SESSION['csrf_token']; ?>" />
     <input type="submit" value="Logout">
   </form>
 <?php else: ?>
   You created a short link! <br />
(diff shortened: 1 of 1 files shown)


examples/url-shortener.php
```diff
@@ -112,33 +112,32 @@
...
 ?>
 
 <div class="box">
 <?php if (isset($authUrl)): ?>
   <div class="request">
-    <a class='login' href='<?= $authUrl ?>'>Connect Me!</a>
+    <a class='login' href='<?php echo $authUrl; ?>'>Connect Me!</a>
   </div>
 <?php elseif (empty($short)): ?>
   <form id="url" method="POST" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
```
tmatsuo: Here is another short tag

examples/url-shortener.php
```diff
@@ -112,33 +112,32 @@
...
-    <input name="url" class="url" type="text">
+    <input name="url" class="url" type="text" placeholder="URL" required>
     <input type="hidden" name="csrf_token" value="<?php echo getCsrfToken() ?>" />
     <input type="submit" value="Shorten">
   </form>
   <form id="logout" method="POST" action="<?= htmlspecialchars($_SERVER['PHP_SELF']); ?>">
```
tmatsuo: Another short tag

Assimilationstheorie:
The Project is full of Short-Tags. I would start to fix it and rebuild the project, but im not sure if this a good idea, because we must rebuild a lot of things! What did you say? If you promise that the project will not be finished now, I will develop it until i die! :D

tmatsuo:
@Assimilationstheorie I don't disagree. How do you want to proceed? Do you want to make another change on this PR and merge this first? or Do you want to file another issue and wait for us to fix them?

Assimilationstheorie:
No Issue, i fix it self. I make another PR and start this week with fix all other files. Okay? :)

bshaffer:
Hello, I am sorry you are not happy with the samples. But it is not necessary to remove short tags. They are 100% okay. You may be confusing them with these tags: `<?`, which were removed entirely in PHP 7.0. Yes, `<?` conflicts with XML and should always be disabled, but `<?=` do not, and can be used safely since PHP 5.4 (they've always been able to be used safely, but before PHP 5.4, they were bundled with `<?` in the same INI configuration so both had to be disabled together). 

As this repo is marked as supported for PHP 5.4 and above, I do not want to change the short tags.

I see that you are also adding a placeholder for the URL, and a `required` tag to the input. This is nice. If you'd like to submit a separate PR with those changes, or update this PR to remove your changes to the short tags, we are happy to merge it!

Assimilationstheorie:
Hey @bshaffer 

Short PHP-Tags are configured in php.ini -> "short_open_tag". Any Configuration in this File are unique and not the same for all people. When you use the "long tag", this will work on all configurations because thats the "default way". We eliminate a potential Error! :)

bshaffer:
@Assimilationsth thank you for your concern but the `short_open_tag` directive was removed in PHP 7.0, and was separated to not include `<?=` in PHP 5.4. So the only people this would cause a potential error for is people using <= PHP 5.3, which is not supported by this library.

Assimilationstheorie:
@bshaffer - Ahh! I see! Never stop learning! :) Thanks for this info!

I will change the short-tags and update the PR! :)

PR #1451, opened 2018-06-12
Replace all PHP-Short-Tags

Assimilationstheorie:


bshaffer:
Same with #1454, there is no reason to replace `<?=` short tags.

Assimilationstheorie:
Same with #1439 :)

PR #1454, opened 2018-06-12
Raplce short PHP-Tags

Assimilationstheorie:


bshaffer:
Why replace PHP short tags? `<?` has been fully removed, but as of PHP 5.4 `<?=` is okay.

Assimilationstheorie:
Same with #1439 :)

PR #1457, opened 2018-06-13
Update simple-upload.

Assimilationstheorie:
file_exists() - Check, if file OR directory exists
is_file() - Check, if only file exist

Rebuild to "is_file()", because we check only for exist file and not dir.


examples/simple-file-upload.php
```diff
@@ -80,7 +80,7 @@
 if ($_SERVER['REQUEST_METHOD'] == 'POST' && $client->getAccessToken()) {
   // We'll setup an empty 1MB file to upload.
   DEFINE("TESTFILE", 'testfile-small.txt');
-  if (!file_exists(TESTFILE)) {
+  if (!is_file()(TESTFILE)) {
```
tmatsuo: Do you want to do `is_file(TESTFILE)` instead?

PR #1459, opened 2018-06-13
Update simple-file-upload

Assimilationstheorie:
file_exists() - Check, if file OR directory exists
is_file() - Check, if only file exist

Rebuild to "is_file()", because we check only for exist file and not dir.


Assimilationstheorie:
Hey @tmatsuo 

I create a new PR for this. :)

PR #1461, opened 2018-06-19
increment libver to 2.3.0

bshaffer:


bshaffer:
After reviewing all changes that have been made since `v2.2.1`, this release is more appropriately `v.2.2.2` since no enhancements have been made.

PR #1465, opened 2018-06-22
Updates footprint cq phpdoc to include string for $scopes

### LINKED ISSUE #1464: The phpdoc of `client::setScopes` typehints only array
The footprint of `client::setScopes()` accepts only array, but it should be both string and array. This would make it equal to the used `client::addScope`, and fixes warnings (e.g. during static analysis) in for instance `ClientTest.php`.

For instance, the test uses this code `$client->setScopes("scope2");` which currently is not correct according to the phpdoc.

NanneHuiges:
Using a string is valid as the addscope() function accepts it as well.
This usage is present in the `clientTest.php` file and also in the
example from the quickstart, but currently gives out a warning on static
analysis of code, as it isn't seen as a valid input.

Resolves #1464

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -725,13 +725,13 @@ class Google_Client
   /**
    * Set the scopes to be requested. Must be called before createAuthUrl().
    * Will remove any previously configured scopes.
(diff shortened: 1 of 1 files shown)


trai-chuavo5000:
Là sao vậy

PR #1468, opened 2018-06-25
examples: Add new Text-To-Speech example using an API key.

### LINKED ISSUE #1426: Please add Texttospeech example
Hi, I'm trying to figure out how to use the new Wavenet voices, but I can't find any examples.

Please help

mattwhisenhunt:
Fixes #1426


bshaffer:
Hey Matt! This is a great example, but we really should be directing users to use [`google/cloud-text-to-speech`](https://github.com/GoogleCloudPlatform/google-cloud-php/tree/master/TextToSpeech) and not to this library.

Issue #1487, opened 2018-07-30
exponential backoff - not possible to provide custom retryMap.


Yahav:
So, i've been trying to use the exponential backoff feature and noticed you won't be able to provide a custom retryMap although it was (faultily) implemented.

So, inspecting Runner.php (Google_Task_Runner) i can see that there is a retryMap array
 ```
 /**
   * @var array $retryMap Map of errors with retry counts.
   */
  protected $retryMap = [
    '500' => self::TASK_RETRY_ALWAYS,
    '503' => self::TASK_RETRY_ALWAYS,
    'rateLimitExceeded' => self::TASK_RETRY_ALWAYS,
    'userRateLimitExceeded' => self::TASK_RETRY_ALWAYS,
    6  => self::TASK_RETRY_ALWAYS,  // CURLE_COULDNT_RESOLVE_HOST
    7  => self::TASK_RETRY_ALWAYS,  // CURLE_COULDNT_CONNECT
    28 => self::TASK_RETRY_ALWAYS,  // CURLE_OPERATION_TIMEOUTED
    35 => self::TASK_RETRY_ALWAYS,  // CURLE_SSL_CONNECT_ERROR
    52 => self::TASK_RETRY_ALWAYS   // CURLE_GOT_NOTHING
  ];
```
as well as a setter(setRetryMap).
tracked it back to REST.php (Google_Http_REST), at the execute function the 5th param is $retryMap which defaults to null, at the function it states that if the provided  $retryMap isn't null, run the setRetryMap() method of the runner, so far so good.
so tracked it back to Client.php (Google_Client) which you suppose to initiate at your code:
`$this->client = new Google_Client();`
the Google_Client class will receive a config array at the constructor 
`  public function __construct(array $config = array())`
the array will then be merged with the default config array.
the default config array contain a key names 'retry' with the following comment
```
// Task Runner retry configuration
// @see Google_Task_Runner
'retry' => array(),
```
So, natural assumption will be that when initiating the Google_Client you should provide an array with a retry key containing your desired retry maps. eg:
 `        $this->client = new Google_Client(array('retry'=>$this->retryMap));`
So far so good?
Except this won't work, Lets analyze what's happening:
we initiate the Google_Client class and provide an array with the a retry key:
 `        $this->client = new Google_Client(array('retry'=>$this->retryMap));`
this array gets merged with the Google_Client's config variable, so now our desired retryMap is stored at the retry key of the class config variable.
[...]

PR #1488, opened 2018-07-30
Fixing the ability to provide custom retryMap

Yahav:
https://github.com/google/google-api-php-client/issues/1487

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -797,11 +797,11 @@ class Google_Client
     $http = $this->authorize();
 
     return Google_Http_REST::execute(
-    $http,
-    $request,
-    $expectedClass,
-    $this->config['retry'],
-    $this->config['retry_map']
+        $http,
+        $request,
+        $expectedClass,
+        $this->config['retry'],
+        $this->config['retry_map']
     );
     
   }
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -130,7 +130,7 @@ class Google_Client
           // Task Runner retry configuration
           // @see Google_Task_Runner
           'retry' => array(),
-          'retryMap' => null,
+          'retry_map' => null,
 
           // cache config for downstream auth caching
           'cache_config' => [],
@@ -796,7 +796,7 @@ class Google_Client
     // this is where most of the grunt work is done
     $http = $this->authorize();
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'],  $this->config['retryMap']);
+    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'], $this->config['retry_map']);
   }
 
   /**
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -796,7 +796,14 @@ class Google_Client
     // this is where most of the grunt work is done
     $http = $this->authorize();
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'], $this->config['retry_map']);
+    return Google_Http_REST::execute(
+    $http,
+    $request,
+    $expectedClass,
+    $this->config['retry'],
+    $this->config['retry_map']
+    );
+    
   }
 
   /**
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -130,6 +130,7 @@ class Google_Client
           // Task Runner retry configuration
           // @see Google_Task_Runner
           'retry' => array(),
+          'retryMap' => null
 
           // cache config for downstream auth caching
           'cache_config' => [],
@@ -795,7 +796,7 @@ class Google_Client
     // this is where most of the grunt work is done
     $http = $this->authorize();
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, null, $this->config['retry']);
+    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'],  $this->config['retryMap']);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Yahav:
I signed it!

src/Google/Client.php
```diff
@@ -795,7 +795,7 @@ public function execute(RequestInterface $request, $expectedClass = null)
     // this is where most of the grunt work is done
     $http = $this->authorize();
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry']);
+    return Google_Http_REST::execute($http, $request, $expectedClass, null, $this->config['retry']);
```
dwsupplee: Thanks for the PR @Yahav :)

From what I can tell, it looks like the `retry` key of the client configuration should map [here](https://github.com/google/google-api-php-client/blob/master/src/Google/Http/REST.php#L44), so the original code looks correct. In order to provide custom retry maps, I believe what we need to do is expose a `retryMap` setting in the `Google_Client` constructor which we can pass down into `Google_Http_REST` like so:

```php
return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'], $this->config['retryMap']);
```

WDYT?

Yahav: The original code isn't working. please read the following carefully:
https://github.com/google/google-api-php-client/issues/1487
the client config 'retry' is documented as:

```
 // Task Runner retry configuration
          // @see Google_Task_Runner
          'retry' => array(),
```

however, no retry config is used at the Google_Task_Runner, there is a 'retries' key but not retry.
this:

```
return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'], $this->config['retryMap']);
```

won't work as an array is expected..
It really seem like the whole task runner wasn't implemented correctly since basic functionality is missing, for instance, you can't define custom retryMap nor can you set the $maxAttempts (which defaults to 1 so basically even the default retryMap won't function correctly since 

```
 /**
   * Checks if a retry can be attempted.
   *
   * @return boolean
   */
  public function canAttempt()
  {
    return $this->attempts < $this->maxAttempts;
  }
```

it will always return true when the maxAttempts is 1, and lacking a way to set it..makes it useless.
the following was written in order to set maxAttempts using a config var but it was never passed.

```
if (isset($config['retries'])) {
      if ($config['retries'] < 0) {
        throw new Google_Task_Exception(
            'Task configuration `retries` must not be negative.'
        );
      }
      $this->maxAttempts += $config['retries'];
    }
```

Anyway, its quite a mess, i'm gonna give it a try and adjust whatever i can to make it functional, i'm also testing this in realtime..

Yahav: Done, should be alright now.  can you confirm?

src/Google/Client.php
```diff
@@ -795,7 +796,7 @@ public function execute(RequestInterface $request, $expectedClass = null)
     // this is where most of the grunt work is done
     $http = $this->authorize();
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry']);
+    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry'], $this->config['retry_map']);
```
dwsupplee: Confirmed this does the trick.

LGTM, except for one tiny nit. This is now breaking the line length guidelines (my fault, it was the way I formatted the code in my suggestion). Could we break this apart over some newlines?

```php
return Google_Http_REST::execute(
    $http,
    $request,
    $expectedClass,
    $this->config['retry'],
    $this->config['retryMap']
);
```

Yahav: Done

src/Google/Client.php
```diff
@@ -795,7 +796,14 @@ public function execute(RequestInterface $request, $expectedClass = null)
...
 
-    return Google_Http_REST::execute($http, $request, $expectedClass, $this->config['retry']);
+    return Google_Http_REST::execute(
+        $http,
+        $request,
+        $expectedClass,
+        $this->config['retry'],
+        $this->config['retry_map']
+    );
+
```
dwsupplee: Could we please drop the whitespace here? LGTM otherwise.

Yahav: done

Yahav:
Can someone merge this?

dwsupplee:
@mattwhisenhunt is this okay to merge?

ggmartins091:
Guys, is this right? 

I know this fixes the ability to provide the custom retryMap, but why are we passing `$this->config['retry']`  as the third argument [here](https://github.com/googleapis/google-api-php-client/blob/master/src/Client.php#L897) when in reality it expects the whole config array as can be seen [here](https://github.com/googleapis/google-api-php-client/blob/16c78c5f1a65e88e6efcc314ece9312031018d64/src/Http/REST.php#L52)
and [here](https://github.com/googleapis/google-api-php-client/blob/16c78c5f1a65e88e6efcc314ece9312031018d64/src/Task/Runner.php#L98)?


Even the `retry` index in that array is incorrect, the Runner class [looks](https://github.com/googleapis/google-api-php-client/blob/16c78c5f1a65e88e6efcc314ece9312031018d64/src/Task/Runner.php#L143) for an index called `retries`. But still, we should pass the whole config array.

dwsupplee:
@ggmartins091 it looks correct to me. `$this->config['retry']` is intended to be the configuration array for the `Runner`, as seen [here](https://github.com/googleapis/google-api-php-client/blob/master/src/Client.php#L151-L153), not a specific number of retry attempts.

ggmartins091:
@dwsupplee Oh, my bad. I get it now.

So for me to change the number of retries I would need to do `new Google_Client(['retry' => ['retries' => 3]]);`

Thats what I was missing. Thank you.

PR #1505, opened 2018-08-31
Use double pipes in composer dependency file

adamsilverstein:
use of single pipes is deprecated, see https://github.com/composer/composer/issues/6755#issue-266478234 and https://getcomposer.org/doc/articles/versions.md#caret-version-range-


adamsilverstein:
🎉

PR #1513, opened 2018-09-10
Add Proxy for verifyIdToken method

NVD-R:
Set proxy for easily like that `$client->verifyIdToken($id_token,['proxy' => 'IP:PORT']`


NVD-R:
I signed it!

PR #1514, opened 2018-09-10
Add Proxy for verifyIdToken method

NVD-R:


mattwhisenhunt:
You can already use set a proxy:
```php
$client->setHttpClient(new GuzzleHttp\Client(['proxy'=> $proxy]));
$payload = $client->verifyIdToken($idtoken);
```

PR #1533, opened 2018-10-11
PHP 7.2 compatibility update

kalifg:
Calling `get_class()` without a parameter is no longer supported in PHP 7.2: https://secure.php.net/manual/en/function.get-class.php


kalifg:
I signed it!

mattwhisenhunt:
Thanks for the PR. I'm going to close it because the change to PHP 7.2 was about passing NULL to the function and in fact encourages use of get_class() without any argument.

https://secure.php.net/manual/en/migration72.incompatible.php#migration72.incompatible.no-null-to-get_class

kalifg:
Sorry about that, I misread the error!

Issue #1552, opened 2018-11-24
Does not clear Batch requests array


daniel55411:
Hello!

I think that after `Batch::execute` need to clear array `Batch::requests`. It will avoid re-creating `Batch` object after executing `Batch::execute` method if batch mode is neccessary to use again.

I know about this [issue](https://github.com/googleapis/google-api-php-client/issues/624) and [pull request](https://github.com/googleapis/google-api-php-client/pull/623), but as far as I understand issue was closed but pull request was not merged.

PR #1554, opened 2018-11-24
In Batch class added clearing requests array after executing

daniel55411:
Fixs #1552


jdpedrie:
I'm on the fence here. This would change behavior, which in a legacy library such as this is dangerous, but I do see how one would expect that it would clear the queue after sending.

However, there are existing (and simple) workarounds, chiefly simply calling `$service->createBatch()` again to create a new, clean batch.

Perhaps improved documentation is a better solution here.

@bshaffer @dwsupplee any thoughts?

PR #1574, opened 2019-01-09
Avoid throwing the exception on unlimited retries

sawmurai:
This fixes an issue where unlimited retries are impossible because -1 is an implicit false.


sawmurai:
I signed it!

danielgsims:
Hi @sawmurai! Could you first provide a unit test that fails the problem. It should also help prove that your change fixes any underlying problems. Thanks!

sawmurai:
Hm actually I think I just misunderstood the code. I think this PR can be closed then.

PR #1621, opened 2019-04-25
Fix usage of undefined constants

m6w6:


dwsupplee:
Thanks for the PR @m6w6 - however,  the existing behavior is correct :). There are some [string interpolation examples here](https://php.net/manual/en/language.types.string.php#language.types.string.parsing.simple) to take a look at.

egiberti:
@dwsupplee While correct today, use of unquoted strings is [deprecated as of PHP 7.2](https://www.php.net/manual/en/migration72.deprecated.php#migration72.deprecated.unquoted-strings) and the `E_NOTICE` is now an `E_WARNING`. This causes bloat in logs and will break in the future.

PHP versions >= 7.2
```
Warning: Use of undefined constant key - assumed 'key' (this will throw an Error in a future version of PHP) in /in/N9kuM on line 7 Interpolation value
```

PHP versions < 7.1
```
Notice: Use of undefined constant key - assumed 'key' in /in/N9kuM on line 7 Interpolation value
```

See https://3v4l.org/N9kuM to test yourself

dwsupplee:
These lines of code don't use brackets in the interpolation. This is a more representative snippet: https://3v4l.org/UMCP3

egiberti:
Interesting that you can do it without curly braces without raising an error... but your code uses curly braces so I'm confused?

https://github.com/googleapis/google-api-php-client/blob/master/src/Google/Http/MediaFileUpload.php#L320

dwsupplee:
Ah, I totally scanned past the brackets in the code 🤦‍♂️. Sorry about that. Yes, you are both correct - reopening! :)

dwsupplee:
@m6w6 When you have a moment if you could please sign the CLA we'll be able to get this merged in :).

m6w6:
Thanks, I signed the CLA.

dwsupplee:
Thanks again @m6w6, we'll have this released early next week.

Issue #692, opened 2015-10-09
use PSR-1 and PSR-2 coding standards


bshaffer:
- update .travis.yml to check CS using `php-cs-fixer`
- ignore `src/Google/Service` as these are generated elsewhere
- http://www.php-fig.org/psr/psr-1/
- http://www.php-fig.org/psr/psr-2/

bshaffer:
Also incorporate PSR-4 by flattening the directory structure

cnelissen:
+1

hakimio:
Any progress?

hacfi:
@bshaffer @jdpedrie Can someone from the community work on this or will this be provided by the Google APIs team as you generate it from some internal template?

bshaffer:
@hacfi We always love and accept community support, but this one may be on us because of the complexity of BC issues. However voicing that you need this really helps with prioritization, and will result in us getting it done. The main reason we haven't taken action on this is nobody's pushed on this from the community.

In [the phpgen-2.0](https://github.com/googleapis/google-api-php-client-services/blob/phpgen-2.0/src/autoload.php) I've proposed an `autoload.php` file that fixes this. In order to move forward, we'd just need to publish a new php generator and add it as a Github action. This is something @jdpedrie and I will hopefully get done this year.

mfn:
FWIF, here's my voice too 🗣️ 

;-)

Genjiff:
Is this still on your roadmap? I would love to see this implemented

repat:
@bshaffer there's definitely some need for this, see https://github.com/googleapis/google-api-php-client/issues/2073 Thanks! :)

PR #1623, opened 2019-04-29
Fix elseif styling based on best practices from PSR2

CodeLingoTeam:
Hi, I've changed a PHP keyword to comply with [this standard](https://www.php-fig.org/psr/psr-2/#51-if-elseif-else) in PSR2. It's admittedly a small fix but I hope it helps!


jdpedrie:
Thanks for the PR. We're tracking this work in #692, which I'll be resurrecting over the next few weeks. So I'm going to wrap this into the larger body of work rather than do it piecemeal.

PR #1626, opened 2019-05-03
Specify configuration filepath  when config file is not found

gpenverne:
It will help to debug


gpenverne:
I signed it!

jdpedrie:
Thanks @gpenverne!

PR #1644, opened 2019-05-29
Add option to toggle API format version

### LINKED ISSUE #1642: Display of detailed error messages
**Current situation.**

When we trying to get some information about some location with Google API, in case of error we receive some general message, like:

```
 {
  "error": {
    "code": 400,
    "message": "Request contains an invalid argument.",
    "errors": [
      {
        "message": "Request contains an invalid argument.",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

Unfortunately, some message does not give us a lot information of debug.
One possible solution: to play with [OAuth 2.0 Playground](https://developers.google.com/oauthplayground/)
**Advantages.** It allows us to get more detail information.
**Disadvantages.** It takes a lot of time to register/prepare request

According to the Google documentation [Support -> Detailed error responses](https://developers.google.com/my-business/content/support#detailed_error_responses)
we can receive full format of answer by adding 'X-GOOG-API-FORMAT-VERSION' => 2  to the Request header.
Problem is that there is no any possibility to add it by passing some input parameters. :(

**Desired solution**
Add possibility to extend Request header by adding some input parameter to the 

`google/apiclient/src/Google/Service/Resource.php -> call()`

something like

`google/apiclient/src/Google/Service/Resource.php -> call($name, $arguments, $expectedClass = null, $isDetailMessage = false)`

and in line 200 except 

`['content-type' => 'application/json'],`

add something like:

`$isDetailMessage ? ['content-type' => 'application/json', 'X-GOOG-API-FORMAT-VERSION' => 2] : ['content-type' => 'application/json']`

or any other solution which you will like more.

After such changes we can receive an error message like:

```
{
  "error": {
    "code": 400,
    "message": "Request contains an invalid argument.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.mybusiness.v4.ValidationError",
        "errorDetails": [
          {
            "code": 1100,
            "field": "address",
            "message": "The specified address cannot be located. Please use the UI to provide a pin drop."
          }
        ]
      }
    ]
  }
}
```

which can save developer's time for debugging:)

Thanks in advance for your time and any information regarding this issue!

jdpedrie:
Fixes #1642.

Usage:
```php
$client = new Google_Client([
    'api_format_v2' => true
]);
```


tests/Google/ClientTest.php
```diff
@@ -684,4 +686,23 @@ public function testTokenCallback()
...
+
+  public function testExecuteWithFormat()
+  {
+    $this->onlyGuzzle6();
+
+    $client = new Google_Client([
+      'api_format_v2' => true
+    ]);
+
+    $guzzle = $this->prophesize('Google_Client');
```
dwsupplee: Can we use `$this->getMock()` here for consistency with the rest of the tests?

jdpedrie: done

dwsupplee: Thanks!

PR #1646, opened 2019-06-04
Delete url-shortener.php

grant:
The URL Shortener API is deprecated and was shut down March 30, 2019.
https://developers.googleblog.com/2018/03/transitioning-google-url-shortener.html

Please review/merge and mark [b/33482850](http://b/33482850) as fixed.

> Requesting review by @bshaffer


grant:
I've updated `examples/index.php` as mentioned.

To resolve the linked bug, would you like to either:
- Keep this sample that uses a discontinued API
or
- Delete this sample

For showing OAuth, perhaps we can cross link the G Suite samples which use OAuth:
https://github.com/gsuitedevs/php-samples

There are ~12 OAuth samples there.

dwsupplee:
Is there a third option of updating it to use another service? I only took a quick glance, but it looks like there is only a single insert call.

bshaffer:
More than happy to link to one of the gsuitedevs samples!

grant:
I added a link in the `examples/` folder to the G Suite samples.

The docs seem to show an example of OAuth with Drive:
https://github.com/googleapis/google-api-php-client#authentication-with-oauth

So I'm not sure exactly why there's the comment of we should be using a different API as Drive already shows OAuth.

grant:
@bshaffer Please review.

bshaffer:
@grant tests are failing. Change [this line](https://github.com/googleapis/google-api-php-client/blob/master/tests/examples/indexTest.php#L29) to look for `8` instead of `9`.

grant:
Done.
That's a super fragile test...

PR #1678, opened 2019-07-16
Remove note about PHP 5.2 support

jdpedrie:
This change removes the part of the README explaining why we still support PHP 5.2, since... we don't, and haven't since [2015](https://github.com/googleapis/google-api-php-client/commit/48dfd9571854d35b2fda1db15b918e6976d7d1e6). 😆


dwsupplee:
Technically we still support the 1.x set of releases, so I think this message still applies. WDYT?

jdpedrie:
How about something instead like "The actively maintained version of the library requires PHP 5.4 or higher. If you require support for PHP 5.2 or 5.3, use v1 of the library"?

dwsupplee:
Sounds good!

jdpedrie:
Done; PTAL

PR #1684, opened 2019-07-29
Run tests against PHP 7.3+ and check for cross-version support for PHP 5.4+

swissspidy:


jdpedrie:
Thanks @swissspidy, the proposal makes a lot of sense! We'll need to address the failing tests, but once that gets ironed out I'm good with this.

swissspidy:
Thanks @jdpedrie! Not sure why the tests fail on PHP 5.4 though 🤔 
I suppose it's because of different versions of packages being used. Any hint appreciated.

swissspidy:
Working smoothly now! 🎉

swissspidy:
BTW I also might have a small change ready that could help with the PHP 7.4 failures. Basically, it would run tests on that platform using a newer version of PHPUnit.

jdpedrie:
Looks good to me. Will `7.4snapshot` still be available once 7.4 is officially released?

swissspidy:
In my experience the snapshot builds on Travis would then just be for the 7.4.x alphas. Because for PHP 8 there is the `nightly` version.

jdpedrie:
So once we get to release, we'll need to be sure to update this to `7.4`.

swissspidy:
Ideally, yes :-)

PR #1685, opened 2019-07-30
Fix typo

vikashsp:


vikashsp:
@googlebot I signed it!

jdpedrie:
https://www.merriam-webster.com/dictionary/abhor

I am of the opinion that "abhor" is not the right word for it though. If you'd like to change it to "If you prefer not to use composer..." I'd be glad to accept that change!

jdpedrie:
See #1697.

PR #1710, opened 2019-09-11
tests: Upgrade unit tests for PHP 7.4 compat

jdpedrie:
In older versions of PHPUnit, the mocking framework has fairly major compatibility issues with PHP 7.4, and due to (in my opinion) questionable decisions taken by the maintainers to not backport fixes to anyone using less than PHP 7.2, will not be taking action to address this in a way we can make use of.

This change replaces the PHPUnit mocks with prophecy mocks. ~~Once [this PR](https://github.com/phpspec/prophecy/pull/432) is tagged by prophecy, we will be able to remove the allow failures tag from PHP 7.4.~~ done.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - 7.4snapshot
+  - 7.4
 
 # Test lowest dependencies on PHP 5.4
 # (Guzzle 5.2, phpseclib 0.3)
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,6 @@ matrix:
   include:
     - php: 5.4
       env: COMPOSER_CMD="composer update phpseclib/phpseclib guzzlehttp/guzzle guzzlehttp/psr7 --prefer-lowest" RUN_PHP_CS=true
-  allow_failures:
-    - php: 7.4snapshot
 
 before_install:
   - composer self-update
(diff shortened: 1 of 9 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
         "guzzlehttp/psr7": "^1.2"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.8.36",
+        "phpunit/phpunit": "^4.8|^5.0",
```
derrabus:
```suggestion
        "phpunit/phpunit": "^4.8.36|^5.6",
```

We need a PHPUnit version that already has the namespaced `TestCase` class alias.

negoziator:
@google-admin  Bump!

.travis.yml
```diff
@@ -24,7 +24,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - 7.4snapshot
+  - 7.4
```
negoziator: :+1:

tests/Google/AccessToken/RevokeTest.php
```diff
@@ -23,128 +23,133 @@
...
-
-    // Test with token string.
-    $revoke = new Google_AccessToken_Revoke($http);
-    $t = $accessToken;
-    $this->assertTrue($revoke->revokeToken($t));
-    $this->assertEquals($accessToken, $token);
-  }
+    public function testRevokeAccessGuzzle5()
+    {
+        $this->onlyGuzzle5();
```
bshaffer: Why did you fix the spaces here but not in the other tests? Is it because you realized that arduous task would crush your soul?

FWIW I think this is great, and if we want we can have a follow-up PR for _just_ converting 2 spaces to 4 spaces

jdpedrie: yes, that's a fairly good description of what happened.

PR #1731, opened 2019-10-24
Remove composer self-update

BackEndTea:
Travis already runs this command by default, so no need to run it twice

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,6 @@ matrix:
   allow_failures:
     - php: 7.4snapshot
 
-before_install:
-  - composer self-update
-
 install:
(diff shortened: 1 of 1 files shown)


jdpedrie:
Since we support old versions of PHP on old travis runtimes, it's hard to guarantee that the environments are updated on the schedule mentioned in the [documentation](https://docs.travis-ci.com/user/languages/php/#installing-composer-packages):

> Note that we update composer every time we update the PHP build environment, which is every 30-60 days.

It's very little overhead to keep running `self-update`, and guards against potential issues. Therefore, we will keep it in place for the immediate future.

Regardless of this decision, I appreciate your contributions! Thank you.

BackEndTea:
Thank you for the explanation on why this won't be merged.

PR #1745, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/examples/idtoken.php
+++ b/examples/idtoken.php
@@ -63,7 +63,7 @@ if (isset($_GET['code'])) {
   $_SESSION['id_token_token'] = $token;
 
   // redirect back to the example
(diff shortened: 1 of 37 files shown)


jdpedrie:
Hi,

We cannot consider this without a signed CLA. Additionally, I don't think this is needed. If you'd like us to reconsider, please sign the CLA as directed by the above comment and provide more information as to the necessity of this change.

Thanks!

PR #1751, opened 2019-12-03
docs: remove shutdown G+ API reference

grant:


README.md
```diff
@@ -2,7 +2,7 @@
 
 # Google APIs Client Library for PHP #
 
-The Google API Client Library enables you to work with Google APIs such as Google+, Drive, or YouTube on your server.
+The Google API Client Library enables you to work with Google APIs such as Cloud Storage, Drive or YouTube on your server.
```
jdpedrie: Could we choose a different service than Storage? Generally we suggest people use `google/cloud-storage` for GCS, as it's a much better user experience than this client. I think calling it out specifically here might confuse that a bit.
```suggestion
The Google API Client Library enables you to work with Google APIs such as Gmail, Drive or YouTube on your server.
```

PR #1762, opened 2019-12-16
Check if the param is an array before trying to access it

negoziator:
Sometimes the request's parameters will end up with the last element is an integer.
That will result in a PHP error.

You cannot access an array with a type Integer

Tested on: PHP 7.4

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Google APIs Client Library for PHP #
 
-The Google API Client Library enables you to work with Google APIs such as Google+, Drive, or YouTube on your server.
+The Google API Client Library enables you to work with Google APIs such as Gmail, Drive or YouTube on your server.
 
 These client libraries are officially supported by Google.  However, the libraries are considered complete and are in maintenance mode. This means that we will address critical bugs and security issues but will not add any new features.
 


negoziator:
@googlebot I signed it!

jdpedrie:
Thanks for the contribution @negoziator. Could you provide an example of where this is causing an error? I'm rather surprised it hasn't been seen before.

src/Google/Service/Resource.php
```diff
@@ -272,6 +272,9 @@ public function createRequestUri($restPath, $params)
     $uriTemplateVars = array();
     $queryVars = array();
     foreach ($params as $paramName => $paramSpec) {
+      if (gettype($paramSpec) !== 'array') {
```
jdpedrie: I'm interested to see that you started with `is_array` before switching. What was the issue with `is_array`?

negoziator: Personally i like the `is_array` over the `gettype`, but i actually didn't think that `is_array` was compatible with `php 5.4`

I can change it back to `is_array`

negoziator:
@jdpedrie  I closed this PR, since i did some digging. I found out that it was actually the "Colors" class that have a fun unexpected null behavior.

I'll open another PR, that takes care of the root-cause instead.

negoziator:
@jdpedrie I've opened another PR :) 

https://github.com/googleapis/google-api-php-client-services/pull/188

src/Google/Service/Resource.php
jdpedrie: `is_array` has been available since PHP 4. Definitely prefer it over `gettype`.

jdpedrie:
Copying my comment from the other pull request since I had to close it. :(

----

@negoziator unfortunately, this repository contains generated code and can't be modified directly. I'm curious about this though:

>  PHP will default the empty array to null

When I run similar code, I don't see it casting anything to null:

```php
function foo($arg = []) {
    $other = [];
    print_r(array_merge($other, $arg));
}

foo();

// Array
// (
// )
```

I think there might be an issue a bit deeper which we need to keep looking for.

If it happens that there is a generator problem, we can work on addressing it there, but I'm not entirely sure that's the case yet.

Let's continue the discussion and research in the original PR since I think it's likely a fix will originate there. :)

negoziator:
@jdpedrie 

So i created a snippet here, where i try to explain the issue.
The idea is that when i loop over a lot of elements - the variable might or might not be initialized to `null` - which gives me the unexpected behavior.

See the snippet here, and let me know what you think :) 

https://web.tinkerwell.app/#/snippets/fe2fc249-1a4e-4ed8-a005-42134f5bd7ac

negoziator:
@jdpedrie ping :)

jdpedrie:
Ideally the generator should be updated to typehint `array $optParams`. That should resolve the problem for you, right?

Why not avoid sending `null` to the `$optParams` argument? It's incorrect usage even if it's technically accepted by the signature. Could you check for null and replace it with an empty array?

PR #1768, opened 2019-12-19
Feat: Add post body to log output

bronhy:
When debugging I would like to see the post body as well.


bronhy:
@googlebot I signed it!

bronhy:
@googlebot I signed it!

PR #1769, opened 2019-12-19
Feat: Add post body to log output.

bronhy:
When debugging i want to see postbody


bronhy:
Hm well that would be a way to approach this but somehow i believe the client config should not have this responsibility. 

If you would like to have his optional i would suggest we create monolog processors. 
The default one would not log the postbody but we can pass in our own which will add the extra context values.

What do you think?

jdpedrie:
I'm fine with that so long as we maintain backwards compatibility in the public API and support >= PHP 5.4.

jdpedrie:
Please feel free to re-open if you'd like to continue working on this! :)

PR #1783, opened 2020-02-04
docs: use explicit getter in sample over iterator interface methods

dwsupplee:
Related to: https://github.com/googleapis/google-api-php-client-services/issues/169#issuecomment-582101676


jdpedrie:
@googlebot I consent.

Issue #1818, opened 2020-04-17
GuzzleHttp 403


tianjincan:
when i request this url using php:
https://www.googleapis.com/androidpublisher/v3/applications/packageName/purchases/products/productId/tokens/token

The json file is downloaded from the Google account, the three parameters are already configured when actually requested，so who can tell me what is wrong,Thank you very much

this is code:
```
$dir = __DIR__."/../../../docs/service-account.json";
        putenv('GOOGLE_APPLICATION_CREDENTIALS='.$dir);

        $packageName = "xxxx";
        $productId = "xxxx";
        $token = "xxxx";
        $client = new Google_Client();
        $client->useApplicationDefaultCredentials();
        $url = "https://www.googleapis.com/androidpublisher/v3/applications/$packageName/purchases/products/$productId/tokens/$token";
        $client->addScope(Google_Service_Plus::PLUS_ME);

        // returns a Guzzle HTTP Client
        $httpClient = $client->authorize();

        // make an HTTP request
        $response = $httpClient->get($url);
        print_r($response);
```
this is resopnse:
```
GuzzleHttp\Psr7\Response Object
(
    [reasonPhrase:GuzzleHttp\Psr7\Response:private] => Forbidden
    [statusCode:GuzzleHttp\Psr7\Response:private] => 403
    [headers:GuzzleHttp\Psr7\Response:private] => Array
        (
            [WWW-Authenticate] => Array
                (
                    [0] => Bearer realm="https://accounts.google.com/", error="insufficient_scope", scope="https://www.googleapis.com/auth/androidpublisher"
                )

            [Vary] => Array
                (
                    [0] => X-Origin
                    [1] => Referer
                    [2] => Origin,Accept-Encoding
                )

            [Content-Type] => Array
                (
                    [0] => application/json; charset=UTF-8
                )

            [Date] => Array
                (
                    [0] => Fri, 17 Apr 2020 02:18:24 GMT
                )

            [Server] => Array
                (
                    [0] => ESF
                )

            [Cache-Control] => Array
                (
                    [0] => private
                )

            [X-XSS-Protection] => Array
                (
                    [0] => 0
                )

            [X-Frame-Options] => Array
                (
                    [0] => SAMEORIGIN
                )

            [X-Content-Type-Options] => Array
                (
                    [0] => nosniff
                )

            [Accept-Ranges] => Array
                (
                    [0] => none
                )

            [Transfer-Encoding] => Array
                (
                    [0] => chunked
[...]

Issue #1701, opened 2019-09-02
Prevent installation of all API Services.


dmzoneill:
10000+ files
dmzoneill@dave-laptop:~/Desktop/ashtangayoga.ie/gapi$ du -sh .
51M	.

Is there a simpler way to do this.  It seems like there is an excessive amount of code to make a few http requests.  none of which i would want to point up on a webserver.  Attack surface....

is the oath flow and response simplified somewhere in this solution or you could recommend something else as reference?

Thanks

jdpedrie:
The files installed by the `google/apiclient-services` dependency are not required for the library to function. They are only required if you need to make calls to that API. If you desire to use this library only for authentication, you may use the `provide` property in your composer file to skip installation of the services:

```json
{
    "provide": {
        "google/apiclient-services": "*"
    }
}
```

jdpedrie:
I'm going to close this issue out. If you're still having trouble, please feel free to re-open!

PR #1823, opened 2020-04-21
docs: update readme with info on services repo

jdpedrie:
This change adds a warning to users to pin to a specific version of `google/apiclient-services` and also a guide to how they can skip installation of services if they don't need them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,8 +47,8 @@ If you're installing this library for authentication or other purposes which do 
 
 ```json
 {
-    "provide": {
-        "google/apiclient-services": "*"
+    "replace": {
+        "google/apiclient-services": "self.version"
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -41,6 +41,18 @@ Finally, be sure to include the autoloader:
 require_once '/path/to/your-project/vendor/autoload.php';
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,18 @@ Finally, be sure to include the autoloader:
 require_once '/path/to/your-project/vendor/autoload.php';
 ```
 
+This library relies on `google/apiclient-services`. That library provides up-to-date API wrappers for a large number of Google APIs. In order that users may make use of the latest API clients, this library does not pin to a specific version of `google/apiclient-services`. **In order to prevent the accidental installation of API wrappers with breaking changes**, it is highly recommended that you pin to the latest version yourself prior to using this library in production.
+
+If you're installing this library for authentication or other purposes which do not require the API wrappers, you may prevent their installation by adding the following to your project's `composer.json` file:
```
bshaffer, line 46: This is confusing to me, as it does not actually prevent their installation. Also, what is the use case for preventing their installation?

jdpedrie: Sorry, should have been `replace` not `provide`. There are a lot of examples of people using this library for convenience in authenticating HTTP clients (e.g. #1818, #1701) without actually using the services.

veperr:
If this is wrong place for this comment/question, sorry, it seemed to close to what I needed, so....
In the install instructions and the text of this issue it says _"**In order to prevent the accidental installation of API wrappers with breaking changes**, it is highly recommended that you pin to the latest version yourself prior to using this library in production."_ But there's no explanation (references, links, etc.) on how to do this. Can this be added?
Also, can you simply give me a link, or some other clue as to how to do this?

bshaffer:
@veperr You can find the latest release of the `google/apiclient-services` package here:

https://github.com/googleapis/google-api-php-client-services/releases

Then you can pin to the latest version by running the following command:

```
composer require google/apiclient-services:0.128
```

PR #1829, opened 2020-05-08
feat: adds new client options

bshaffer:
Depends on https://github.com/googleapis/google-auth-library-php/pull/277 for the `quota_project` option to work


jdpedrie:
Probably want to get auth tagged with the dependent change before merging this.

PR #1836, opened 2020-05-18
chore: Move from Travis to GitHub Actions

bshaffer:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,79 @@
...
+        runs-on: ${{matrix.operating-system}}
+        strategy:
+            matrix:
+                operating-system: [ ubuntu-latest ]
+                php: [ "5.6", "7.0", "7.1", "7.2", "7.3", "7.4" ]
+        name: PHP ${{matrix.php }} Unit Test
+        steps:
+            - uses: actions/checkout@v2
+            - name: Setup PHP
+              uses: nanasess/setup-php@v3.0.4
```
jdpedrie, line 19: I've found that `shivammathur/setup-php@v2` is way faster, especially for early 5.x versions. It's a drop-in replacement.

PR #1852, opened 2020-06-18
chore: automate asset release

jdpedrie:
This adds a Github action to automatically generate and upload release assets when a new release is created.

This has a couple advantages over the manual process:
- It runs natively on multiple versions of PHP, so we don't need `--prefer-lowest`.
- We can generate releases for more versions of PHP. I'm running it here for 5.4, 5.6, 7.0 and 7.4.
- **It's automatic**

[Sample Run](https://github.com/jdpedrie/google-api-php-client/runs/784602695)
[Sample Release](https://github.com/jdpedrie/google-api-php-client/releases/tag/v2.4.15)

### DIFF
--- a/.github/workflows/asset-release.yml
+++ b/.github/workflows/asset-release.yml
@@ -1,0 +1,61 @@
+name: Add Release Assets
+
+on:
+    release:
+        types: [published]
+
+jobs:
+    asset:
+        runs-on: ${{ matrix.operating-system }}
+        strategy:
+            matrix:
+                operating-system: [ ubuntu-latest ]
+                php: [ "5.4", "5.6", "7.0", "7.4" ]
+
+        name: Upload Release Assets
+        steps:
+            - uses: olegtarasov/get-tag@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/asset-release.yml
```diff
@@ -0,0 +1,61 @@
...
+                timeout_minutes: 10
+                max_attempts: 3
+                command: composer remove --dev cache/filesystem-adapter && composer install --no-dev --prefer-dist
+
+            - name: Create Archive
+              run: |
+                zip -r ${fileName} . &&
+                zip -d ${fileName} ".git*" &&
+                zip -d ${fileName} "tests*" &&
+                zip -d ${fileName} "examples*"
```
bshaffer, line 49: Do we also want to delete `docs`, `.travis.yml`, `php.cs.dist` and `.phpunit.xml.dist`?

jdpedrie: Good thinking. done.

PR #1860, opened 2020-06-28
Support Guzzle7

Nyholm:
I did a quick search and found a few places where I could fix stuff. 
I have not found any other incompatibility issues, but this is not manually tested. 

The upgrade from Guzzle 6 to 7 is really smooth. See [upgrade docs](https://github.com/guzzle/guzzle/blob/master/UPGRADING.md). But the upgrade to a future Guzzle 8 will not be. 

I strongly recommend to move towards PSR-18. Guzzle 7 support PSR-18 out of the box. 

Note: #1861

Blocked by https://github.com/googleapis/google-auth-library-php/pull/256

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "firebase/php-jwt": "~2.0||~3.0||~4.0||~5.0",
         "monolog/monolog": "^1.17|^2.0",
         "phpseclib/phpseclib": "~0.3.10||~2.0",
-        "guzzlehttp/guzzle": "~5.3.1||~6.0",
+        "guzzlehttp/guzzle": "~5.3.1||~6.0||~7.0",
         "guzzlehttp/psr7": "^1.2"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -5,9 +5,12 @@
     "keywords": ["google"],
     "homepage": "http://developers.google.com/api-client-library/php",
     "license": "Apache-2.0",
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/jeromegamez/google-auth-library-php"}
+    ],
     "require": {
         "php": ">=5.4",
-        "google/auth": "^1.9",
+        "google/auth": "dev-guzzle7 as 1.10",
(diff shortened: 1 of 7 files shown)


jdpedrie:
@Nyholm out of curiosity, is there a roadmap and target date for Guzzle 8?

src/Google/Client.php
```diff
@@ -1124,22 +1124,29 @@ public function setApiFormatV2($value)
...
+        $options['handler'] = new StreamHandler();
         $options['defaults']['verify'] = '/etc/ca-certificates.crt';
       }
-    } else {
-      // guzzle 6
+    } elseif (6 === $guzzleVersion || 7 === $guzzleVersion) {
+      // guzzle 6 or 7
       $options['base_uri'] = $this->config['base_path'];
+    } else {
+      throw new LogicException('Could not find support version of Guzzle.');
```
jdpedrie, line 1149:
```suggestion
      throw new LogicException('Could not find supported version of Guzzle.');
```

composer.json
```diff
@@ -5,14 +5,17 @@
     "keywords": ["google"],
     "homepage": "http://developers.google.com/api-client-library/php",
     "license": "Apache-2.0",
+    "repositories": [
+        { "type": "vcs", "url": "https://github.com/jeromegamez/google-auth-library-php"}
+    ],
     "require": {
         "php": ">=5.4",
-        "google/auth": "^1.9",
+        "google/auth": "dev-guzzle7 as 1.10",
```
jdpedrie, line 13: need to fix this before merging.

Nyholm:
Excellent question. 
There is no written down plan except for the 8.0 github milestone with some issues and half finished work. 

I know I’ve turned down many PRs because they would make the upgrade path too complicated from 6 to 7. However since Guzzle 7+ is implementing psr18, then I’m okey with having a bit more complex upgrade path between 7 and 8. 

I know we need to clear up our messy exceptions inheritance and get rid of a lot of magic functions.

bshaffer:
added in #1868

Issue #1864, opened 2020-07-02
Guzzle Version 7


sambenne:
Can we get Guzzle to use version 7 as some libraries can't be installed as this package is blocking it.

bshaffer:
This has already been added to `google/auth`, and we just need to tag a version for it to work. I'll be able to make that happen today!

bshaffer:
added in #1868

PR #1865, opened 2020-07-02
#1864 - Add Guzzle version 7

sambenne:
#1864 Added Guzzle version 7 for new packages.


bshaffer:
There is more to it than that. Closing this in favor of #1860

PR #1868, opened 2020-07-08
feat: Guzzle 7 support

bshaffer:
Same as #1860 but updated for `google/auth:1.10.0` and a few other fixes.


Nyholm:
Wohoo. Thank you very much!

PR #1872, opened 2020-07-09
feat: add docs generation action and fix docs

bshaffer:
Documentation generation for `google-api-php-client`

Example: http://bshaffer.github.io/google-api-php-client/master/


.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+        steps:
+        - name: Checkout
+          uses: actions/checkout@v2
+        - run: git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+        - name: Install Dependencies
+          uses: nick-invision/retry@v1
+          with:
+            timeout_minutes: 10
+            max_attempts: 3
+            command: composer config repositories.sami vcs https://${{ secrets.GITHUB_TOKEN }}@github.com/jdpedrie/sami.git && composer require sami/sami:dev-master && git reset --hard HEAD
```
bshaffer, line 22: This should be updated to use a tag once we have one!

PR #1875, opened 2020-07-10
chore: fix deprecated syntax

jakubpas:
Remove depreciation warnings for php 7.4 which cause some unit tests to fail


jakubpas:
@googlebot I signed it!

jakubpas:
Remove depreciation warnings for php 7.4 which cause some unit tests to fail

Issue #1867, opened 2020-07-07
Is it necessary to make this package so big? 


camohub:
I have a small host which counts 100MB. I need your lib to log in. This lib has 26MB. Its few times bigger than other libs including whole framework. It is not possible to decrease the size or divide it to more than one package? I dont need tests nor games libs .....

jdpedrie:
If you are using this client for authentication only, you may be able to accomplish your goals using the [`google/auth`](https://github.com/googleapis/google-auth-library-php) library.

If you're not using any APIs in `google/apiclient-services`, you can prevent it from being installed by adding the following to your root `composer.json` file:

```json
{
	"replace": {
		"google/apiclient-services": "self.version"
	}
}
```

If you're using the downloadable package, or need some but not all of the included clients, you could simply delete the unnecessary files and directories from `vendor/google/apiclient-services/src`.

bshaffer:
To use the `google/auth` package to retrieve an access token for a user, you can do the following:

```php
// Create the auth client
$credenitalsFile = '/path/to/client-credentials.json';
$credentials = json_decode(file_get_contents($credentialsFile), true);
$oauth2 = new Google\Auth\OAuth2([
    'client_id' => $credentialsFile['client_id'],
    'client_secret' => $credentialsFile['client_secret'],
    'scope' => 'https://www.googleapis.com/auth/drive',
]);

// Rudimentary filecache
$accessTokenFile = '/path/to/access-token.json';

if (isset($_GET['code'])) {
    $oauth2->setCode($_GET['code']);
    // If we have a code back from the OAuth 2.0 flow, exchange that for an access token.
    file_put_contents($accessTokenFile, json_encode($oauth->fetchAuthToken()));
}

if (file_exists($accessTokenFile)) {
    // $accessToken has a refresh token because "access_type" is set to "offline"
    $accessToken = json_decode(file_get_contents($accessTokenFile));
    $oauth->updateToken($accessToken);
} else {
    // Redirect the user back to this page after authorization
    $redirectUri = 'http://' . $_SERVER['HTTP_HOST'] . $_SERVER['PHP_SELF'];
    $authUri = $oauth->buildFullAuthorizationUri(['redirect_uri' => $redirectUri]);
    // Redirect the user to the authorization URI
    header('Location: ' . $authUri);
    return;
}

// Make the call with the access token
$middleware = new Google\Auth\Middleware\AuthTokenMiddleware($oauth2);
[...]

PR #1877, opened 2020-07-10
feat: composer cleanup task

bshaffer:
An attempt to solve #1867 for those who have filesize limits:

In your `composer.json`:
```
{
    "require": {
        "google/apiclient": "^2.7"
    },
    "extra": {
        "google/apiclient-services": [
            "YouTube"
        ]
    },
    "scripts": {
        "post-update-cmd": "Google_Task_Composer::cleanup"
    }
}
```

This will remove all services but `YouTube`. If a non-existent service is supplied, an exception is thrown.


src/Google/Task/Composer.php
```diff
@@ -0,0 +1,96 @@
...
+use Composer\Script\Event;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+
+class Google_Task_Composer
+{
+  public static function cleanup(Event $event)
+  {
+    $extra = $event->getComposer()->getPackage()->getExtra();
+    $servicesToKeep = $extra['google/apiclient-services'] ?? [];
```
jdpedrie, line 27: Defaulting this seems a bit unfriendly. If a user fat-fingered something, they'd delete all the services silently.

Additionally, I doubt this will become a problem, but we may as well sub-array the list in case we come up with another reason to use `extra`.

```json
{
	"google/apiclient-services": {
		"keep": [
			"YouTube"
		]
	}
}
```

We could use an empty array as a signal to delete all. If `extra."google/apiclient-services".keep` doesn't exist, raise a warning or error.

jdpedrie: also `??` doesn't work for php 5.

bshaffer: I've added service validation so it *shouldnt* be possible to remove all services

bshaffer: I've added tests to confirm this

tests/Google/Task/ComposerTest.php
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    private function createMockFilesystem(array $files, $serviceDir)
+    {
+        $consecutive = [];
+        foreach ($files as $filename) {
+            $file = new \SplFileInfo($serviceDir . $filename);
+            $consecutive[] = [$this->equalTo($file->getRealPath())];
+        }
+        $mockFS = $this->getMockBuilder('Symfony\Component\Filesystem\Filesystem')
```
jdpedrie, line 102: PHPUnit's Mock Builder isn't really compatible with PHP 7.4. You should use Prophecy. The rest of the unit tests have been migrated already.

src/Google/Task/Composer.php
bshaffer: TBH I am not a fan of the "keep" identifier. I think we can exclude it for now and we could always whitelist keys like that later.

tests/Google/Task/ComposerTest.php
bshaffer: Is Prophecy compatible all the way back to 5.4?

jdpedrie: Yep

jdpedrie:
@bshaffer what's the latest with this?

tests/Google/Task/ComposerTest.php
```diff
@@ -0,0 +1,147 @@
...
+                ->shouldBeCalledTimes(1);
+
+            $mockEvent->getIO()
+                ->shouldBeCalledTimes(1)
+                ->willReturn($mockIO->reveal());
+        }
+
+        return $mockEvent->reveal();
+    }
+}
```
dwsupplee, line 147: One  minor nit, no newline here

bshaffer:
@jdpedrie @dwsupplee the `README` has been updated with instructions. Please take a look as there is an ugly caviat involved.

tests/Google/Task/ComposerTest.php
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        return $mockEvent->reveal();
+    }
+
+    public function testE2E()
+    {
+        $composerJson = json_encode([
+            'require' => [
+                'google/apiclient' => 'dev-add-composer-cleanup'
```
bshaffer, line 152: We will need to change this before tagging a new version!

PR #1885, opened 2020-07-18
chore: added namespaces and psr-4 

ivanfretes:


bshaffer:
Hi @ivanfretes !
This change backwards compatibility so unfortunately we can't merge it.

PR #1895, opened 2020-07-27
Add content-type configurability in Resource->call method

alberto-bottarini:
Some methods, for example `Google_Service_Doubleclicksearch_Resource_Reports.getFile` does not accept `application/json` as content-type of the request.

With this PR, each services could change the accepted content-type.

### DIFF
--- a/src/Google/Service/Resource.php
+++ b/src/Google/Service/Resource.php
@@ -77,7 +77,7 @@ class Google_Service_Resource
    * @return Google_Http_Request|expectedClass
    * @throws Google_Exception
    */
-  public function call($name, $arguments, $expectedClass = null)
+  public function call($name, $arguments, $expectedClass = null, $contentType = 'application/json')
   {
     if (! isset($this->methods[$name])) {
       $this->client->getLogger()->error(
@@ -197,7 +197,7 @@ class Google_Service_Resource
     $request = new Request(
         $method['httpMethod'],
         $url,
-        ['content-type' => 'application/json'],
+        ['content-type' => $contentType],
         $postBody ? json_encode($postBody) : ''
     );
 


alberto-bottarini:
After this merge, you could tag this version, so I can fix `Google_Service_Doubleclicksearch_Resource_Reports.getFile` method in `google-api-php-client-services` project.

Thanks

jdpedrie:
Thanks for the contribution @alberto-bottarini. The code in `google/apiclient-services` is automatically generated, so ideally we'd find a way to make this work without requiring changes to that client, as they'll be overwritten at next generation. As you've discovered, we currently hard-code the content type, which is not going to work for you. I wonder if there's another way to attack the problem.

Here's a way you could get this working now. I'm open to library modifications that make this easier, so long as the fit the existing constraints.

```php
$client->setScopes("https://www.googleapis.com/auth/doubleclicksearch");
$service = new Google_Service_Doubleclicksearch($client);
$client->setDefer(true);

$req = $service->reports->getFile($reportId, $fragment);
$req = $req->withHeader('content-type', 'text/plain');

$res = $client->execute($req);
$client->setDefer(false);
```

alberto-bottarini:
Ok thanks for your reply.
I'm sorry I did not read with careful about code generation.

What is used as source for code generation?

jdpedrie:
https://github.com/google/apis-client-generator

alberto-bottarini:
Just a question, is not possible to update library that generates code in order to add this parameter?
Now it has a sort of bug...

jdpedrie:
Is the snippet I shared before not working?

alberto-bottarini:
Yes but it's a workaround. I would like to to contribute in order to have a
better and more stable api library.

Il lun 24 ago 2020, 21:06 John Pedrie <notifications@github.com> ha scritto:

jdpedrie:
I see. Yes, the API generator is open to contributions!

PR #1914, opened 2020-08-19
fix: use quota_project_id over quota_project

dwsupplee:
Depends upon: https://github.com/googleapis/google-auth-library-php/pull/299

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -1169,7 +1169,7 @@ class Google_Client
         'client_email' => $this->config['client_email'],
         'private_key' => $signingKey,
         'type' => 'service_account',
-        'quota_project' => $this->config['quota_project'],
+        'quota_project_id' => $this->config['quota_project'],
       );
       $credentials = CredentialsLoader::makeCredentials(
           $scopes,


src/Google/Client.php
```diff
@@ -1169,7 +1169,7 @@ private function createApplicationDefaultCredentials()
         'client_email' => $this->config['client_email'],
         'private_key' => $signingKey,
         'type' => 'service_account',
-        'quota_project' => $this->config['quota_project'],
+        'quota_project_id' => $this->config['quota_project'],
```
bshaffer, line 1172: Two thoughts. Would it be wise if we did these items as well?

1. Change `quota_project` option to `quota_project_id` for consistency w/ `google/auth`
1. Set both `quota_project` AND `quota_project_id` in the auth library so that using this client w/ an older version of `google/auth` continues to work

```php
$quotaProject = isset($this->config['quota_project_id'])
    ? $this->config['quota_project_id']
    : isset($this->config['quota_project']) ? $this->config['quota_project'] : null;
      $serviceAccountCredentials = array(
        // ...
        'quota_project' => $quotaProject,
        'quota_project_id' => $quotaProject,
 );
```

It's a little more convoluted, but hopefully over all a better solution. wdyt?

bshaffer: I am more concerned about 1. than 2, as it's dangerous to use such similar but slightly different options names for our libraries

bshaffer: And for 2. we can set minimal `google/auth` version to `^1.12` (once it's released)

dwsupplee: Sounds reasonable to me. 👍 

I was thinking of changing the naming of the parameters and such in auth as well, but then we'd need to change the interface (e.g., getQuotaProject -> getQuotaProjectId) so it seems best to leave them as is - wdyt?

bshaffer: After further investigation, `quota_project` is consistently the option, and `quota_project_id` is internal to the JSON key file, and is only used in reference to the JSON key file. So this is LGTM.

PR #1929, opened 2020-09-10
feat: allow array, null config constructors for services

bshaffer:
Rather than force users to constantly create the `Google_Client` class, create it as part of the `Google_Service` constructor:

```php
// this works!
$service = Google_Service_Drive();

// this also works!
$service = Google_Service_Drive([
    'client_id' => 'my-client',
    'client_secret' => '123abc'
]);

// and of course, this still works...
$service = Google_Service_Drive($client);
```

**NOTE**: because [generated service class constructors](https://github.com/googleapis/google-api-php-client-services/blob/master/src/Google/Service/AbusiveExperienceReport.php#L32) type-hint for `Google_Client`, this change will not affect anything until we update the generator to remove the typehint.


dwsupplee (approved):
I like it 👍

src/Google/Service.php
```diff
@@ -25,9 +25,21 @@ class Google_Service
   public $resource;
   private $client;
 
-  public function __construct(Google_Client $client)
+  /**
+   * @param Google_Client|array $clientOrConfig Optional
+   */
+  public function __construct($clientOrConfig = array())
```
jdpedrie, line 31: nit: 
```suggestion
  public function __construct($clientOrConfig = [])
```

src/Google/Service.php
```diff
@@ -25,9 +25,21 @@ class Google_Service
...
   {
-    $this->client = $client;
+    if ($clientOrConfig instanceof Google_Client) {
+      $this->client = $clientOrConfig;
+    } elseif (is_array($clientOrConfig)) {
+      $this->client = new Google_Client($clientOrConfig ?: array());
```
jdpedrie, line 36: nit:

```suggestion
      $this->client = new Google_Client($clientOrConfig ?: []);
```

src/Google/Service.php
```diff
@@ -25,9 +25,21 @@ class Google_Service
...
+    } else {
+      $errorMessage = 'constructor must be array or instance of Google_Client';
+      throw class_exists('TypeError')
```
jdpedrie, line 39: I'm not sold on this condition, especially since the only way to catch both in a single block is using `Throwable`. Why not just use `InvalidArgumentException`?

bshaffer: I am on the fence as well but figure we should throw the TypeError when we can for two reasons: 1. TypeError is the more correct thing to do and 2. it's less disruptive since this is currently throwing a TypeError.

Willing to change it back though...

jdpedrie: ah, that's a good point. I guess it's unlikely that somebody would be catching a bad constructor argument exception anyhow -- it's almost 100% a bug, right? not a recoverable condition, more something you fix and re-run the script.

Since in PHP 5 such a case would be a fatal error, maybe instead of throwing `InvalidArgumentException` we should trigger a fatal error if `TypeError` is undefined. I think that would maintain current behavior in both major versions of PHP.

bshaffer: I can get behind that. Something like this?

```php
trigger_error($errorMessage, E_USER_ERROR);
```

jdpedrie: Yep

jdpedrie (approved):

PR #1935, opened 2020-09-12
feat: add caching for GCE and access tokens by default

bshaffer:
- Passes cache to calls to `ApplicationDefaultCredentials::getCredentials` to take advantage of new onGce cache
 - Wraps credentials in `FetchAuthTokenCache` on calls to `ApplicationDefaultCredentials::fetchAccessTokenWithAssertion`


bshaffer:
depends on a fix in https://github.com/googleapis/google-api-php-client/pull/1968/files which somehow passed when we added namespaces

jdpedrie (approved):

PR #1937, opened 2020-09-14
feat: rename namespaces and provide aliases

bshaffer:
Addresses #692:
 - Adds namespaces, removes underscores
 - Adds autoloading for namespaces
 - Flattens directories for PSR-4

## Backwards-compatibility concerns
BC is preserved using [`aliases.php`](https://github.com/googleapis/google-api-php-client/blob/phpgen-2.0/src/aliases.php). This loads the (minimal) 20 classes in this library upfront and creates aliases for them.

Another option is to define a custom `autoload.php` which lazy-loads the `Google_` classes as they're requested:

```php
// src/autoload.php
$classMap = [
    'Google_Client' => 'Google\Client',
    // ...
];

spl_autoload_register(function ($class) use ($classMap) {
    if (isset($classMap[$class])) {
        class_alias($classMap[$class], $class);
        return true;
    }
});
```

The problem with this is the `autoload.php` function is invoked on every class-exists. Which is preferred? Tell me your thoughts in the comment section and don't forget to smash that "subscribe" button!

**Note**: this is independent of `google/apiclient-services`


marktheshark-33 (approved):

src/aliases.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+$classMap = [
```
jdpedrie, line 3: I'm worried about this map falling out of date if we add new classes. Would be nice to test that all the classes in the `src` folder within the `Google\\` namespace have a mapping. Some combination of `RecursiveDirectoryIterator` and [`FileReflector`](https://github.com/phpDocumentor/Reflection/blob/3.0/src/phpDocumentor/Reflection/FileReflector.php) maybe.

Also, maybe just my preference, but I'd reverse the keys and values of this array. Makes more sense to me that the real name is the key, and also corresponds better to the argument order of `class_alias()`.

jdpedrie:
I think I have a small preference for the `aliases.php` file over an autoloader, but it may be because it already exists.

src/aliases.php
mfn: > I'm worried about this map falling out of date if we add new classes

Stupid outsider question (apologies if this isn't appropriate here!): why would a _new_ class still be relevant for the old non-namespaced version? Would this still be maintained _in parallel_ … 🤔

Thanks!

jdpedrie: That's a good point, I suppose it's something that we'd need to consider. It would seem strange to have some classes without aliases, but maybe that's good, might be a nudge towards upgrading to use the namespace.

bshaffer: I switched the order of `aliases.php` 

I agree with @mfn that there'd be no reason to add an alias for a new class

bshaffer:
@jdpedrie I think it makes things easier and simpler to autoload `aliases.php` up front than to mess with an autoloader. The number of classes is small enough that it's unlikely to have a noticeable impact.

jdpedrie (approved):

Issue #1964, opened 2020-10-13
Default Http Client for Google_Client has wrong default setting


sylvain-msl:
#### Description

Google Api Client `Google_Client` works with Guzzle Http starting from version 5.0 to version 7.0. Version 7 introduced a deprecation for `exceptions` key in config options. It is currently used in `createDefaultHttpClient` https://github.com/googleapis/google-api-php-client/blob/d0dcc7054faad3cdda5844e15103a59984969744/src/Google/Client.php#L1151

For version 7 and up `http_errors` should be used instead.

This can lead to unexpected behavior, for instance when used in pair with the Google `OAuthClient` that is not catching Guzzle `RequestExceptions`

#### Environment details

  - OS: Linux Debian
  - PHP version: 7.2.28
  - Package name and version: google/apiclient v2.7.2 and guzzlehttp/guzzle 7.0

#### Steps to reproduce

  1. Create a new `Google_Client`
  1. Set `refresh_token` with a non working token
  1. Catch a Guzzle `RequestException` instead of `Google_Service_Exception`

#### Code example

```php
use GuzzleHttp\Exception\RequestException;
use Google_Service_Exception;

try {
    $client = new Google_Client();
    $client->setClientId('an_id');
    $client->setClientSecret('a_secret');
    $client->refreshToken('a_wrong_refresh_token');
} catch (Google_Service_Exception $exception) {
    echo 'google exception';
} catch (RequestException $exception) {
    echo 'not google exception';
}
```

#### Proposed Quick Fix

After https://github.com/googleapis/google-api-php-client/blob/d0dcc7054faad3cdda5844e15103a59984969744/src/Google/Client.php#L1151
add
```php
if (7 === $guzzleVersion && isset($options['exceptions']) {
    $options['http_errors'] = $options['exceptions'];
    unset($options['exceptions']);
}
```

jdpedrie:
Hi @sylvain-msl, thanks for bringing this to our attention. The proposed solution seems good to me. Would you like to open a pull request?

sylvain-msl:
@jdpedrie Sure. will do.

bshaffer:
Fixed in #1966

PR #1966, opened 2020-10-19
Issue 1965 - Exceptions option fallback to http_errors on guzzle 6 or 7

sylvain-msl:
## Misc

Fix for #1964 

### Updates

- default options `exceptions` changed to `http_errors` when guzzle version is 6 or 7. (The deprecation of `exceptions` was implemented in guzzle 6 and enforced in guzzle 7).

### DIFF
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -1162,6 +1162,9 @@ class Google_Client
     } elseif (6 === $guzzleVersion || 7 === $guzzleVersion) {
       // guzzle 6 or 7
       $options['base_uri'] = $this->config['base_path'];
+
+      $options['http_errors'] = $options['exceptions'];
+      unset($options['exceptions']);
     } else {
       throw new LogicException('Could not find supported version of Guzzle.');
     }
--- a/src/Google/Client.php
+++ b/src/Google/Client.php
@@ -1148,11 +1148,10 @@ class Google_Client
       $guzzleVersion = (int)substr(ClientInterface::VERSION, 0, 1);
     }
 
-    $options = ['exceptions' => false];
     if (5 === $guzzleVersion) {
       $options = [
         'base_url' => $this->config['base_path'],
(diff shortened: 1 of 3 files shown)


src/Google/Client.php
```diff
@@ -1162,6 +1162,9 @@ protected function createDefaultHttpClient()
     } elseif (6 === $guzzleVersion || 7 === $guzzleVersion) {
       // guzzle 6 or 7
       $options['base_uri'] = $this->config['base_path'];
+
+      $options['http_errors'] = $options['exceptions'];
+      unset($options['exceptions']);
```
bshaffer, line 1167: Seems like we should just remove the setting of `exceptions` above and do the following instead for more obvious/straightforward logic:

```php
    if (5 === $guzzleVersion) {
      $options = [
        'base_url' => $this->config['base_path'],
        'defaults' => ['exceptions' => false],
      ];
      if ($this->isAppEngine()) {
        // set StreamHandler on AppEngine by default
        $options['handler'] = new StreamHandler();
        $options['defaults']['verify'] = '/etc/ca-certificates.crt';
      }
    } elseif (6 === $guzzleVersion || 7 === $guzzleVersion) {
      // guzzle 6 or 7
      $options[
          'base_uri' => $this->config['base_path'],
          'http_errors' => false,
      ];
    } else {
      throw new LogicException('Could not find supported version of Guzzle.');
    }
```

bshaffer (changes requested):
Thank you for your contribution! I cleaned up the submission a bit to make it more readable w/o changing the functionality. If you can make these changes then we can merge this.

src/Google/Client.php
sylvain-msl: Done. It makes more sense indeed. Thanks for your feedback.

bshaffer (approved):

bshaffer:

PR #1978, opened 2020-10-27
fix: preload errror with aliases.php

### LINKED ISSUE #1976: Aliases might be incompartible with PHP 7.4 Preloading feature
#### Environment details

  - OS: docker php:7.4-fpm
  - PHP version: php:7.4
  - Package name and version: latest (2.8.0)

#### Steps to reproduce

I have an issue with [aliases](https://github.com/googleapis/google-api-php-client/blob/master/src/aliases.php) in Symfony app optimized for PHP 7.4 [preloading feature](https://www.php.net/manual/en/opcache.preloading.php). It loads all the required classes at once and conflicts with the aliases file which leads to numerous errors like

```
Cannot declare class Google_Http_REST, because the name is already in use in /app/vendor/google/apiclient/src/aliases.php on line 27
```

Could you please make it optional or add checking that those classes are not defined yet?

For those who have the same issue possible solutions are
-  stick to 2.7.x (unfortunately not my case because I use it indirectly via lib with dependency like ^2.x)
- in a build script wipe out aliases file with `echo '<?php' > ./vendor/google/apiclient/src/aliases.php`

bshaffer:
Fixes #1976 by not declaring aliases when classes already exist

### DIFF
--- a/src/aliases.php
+++ b/src/aliases.php
@@ -1,5 +1,11 @@
 <?php
 
+if (class_exists('Google_Client')) {
+    // Prevent error with preloading in PHP 7.4
+    // @see https://github.com/googleapis/google-api-php-client/issues/1976
+    return;
+}
+
 $classMap = [
     'Google\\Client' => 'Google_Client',
     'Google\\Service' => 'Google_Service',


bshaffer:
Depends on https://github.com/googleapis/google-api-php-client/pull/1979

dwsupplee (approved):

PR #1984, opened 2020-10-28
Allow static analysis tools and IDEs to understand

BackEndTea:
This should allow IDE's to better understand what is going on with the new clases

### DIFF
--- a/src/aliases.php
+++ b/src/aliases.php
@@ -40,3 +40,26 @@ foreach ($classMap as $class => $alias) {
 class Google_Task_Composer extends \Google\Task\Composer
 {
 }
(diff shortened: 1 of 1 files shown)


bshaffer (approved):
This looks great! But this will need to pass our style-checker. You can exclude `aliases.php` in our style config, or just put the curly braces on a new line (as I did with `Google_Task_Composer`).

bryanlaura736 (approved):

bshaffer:
@BackEndTea Please respond and update this PR accordingly! We'd like to get this submitted, but we need it to pass our style checker!

PR #1991, opened 2020-11-10
Do not use autoload when check class existence.

ilyaguy:
Do not use autoload functionality when check class existence.
With default "true" it will fail on custom autoload.


ilyaguy:
@googlebot I signed it!

bshaffer (approved):

PR #2017, opened 2020-12-14
PHP 8 Support

Nielsvanpach:
Hi there, I've made some adjustments to make this package work with PHP 8!
However,  we need a consensus on the supported php versions before all tests can pass.

**Supported PHP Versions**
For PHP versions < 7.2 I've included a script for making the tests PHPUnit ^6 compatible. Thanks to php-webdriver for this approach: https://github.com/php-webdriver/php-webdriver/pull/828/files#

PHP 5.4 and 5.5 are both still supported now. At the moment they are 4 and 5 years EOL. With the release of PHP 8 last month, I believe it might be time to let them go in the latest version (with PHP 8 support).

You can see the test results on my fork: https://github.com/Nielsvanpach/google-api-php-client/actions/runs/421025986
 Everything passed, except 5.4 and 5.5. As I already said, I suggest to drop support on them. I don't think it's really useful to spend time making the test suite 5.4 and 5.5 compatible.


Nielsvanpach:
@googlebot I signed it!

bshaffer:
/cc @teboring for this approach to patching `phpunit` for PHP 5.6/7.0/7.1 

About [3% of Wordpress](https://wordpress.org/about/stats/) runs on PHP 5.4 and 5.5. These calls likely make up only a fraction of the users of this package. The percentage using [Composer](https://blog.packagist.com/php-versions-stats-2020-2-edition/) is even smaller.

I will need to do some data mining to verify the number of calls currently being made to our APIs from these PHP versions to affirm this is acceptable, but I suspect it will be. 

@Nielsvanpach This is a great PR and I appreciate your contribution! If you'd like, you can remove PHP 5.4 and 5.5 from the asset release and test workflows in this PR as well, and update `composer.json` minimum requirement. 

Dropping support for PHP 5.4 and 5.5 will not require a major version bump. As updating minimum PHP requirement ensures existing installations on PHP 5.4 and 5.5  will not break.

Nielsvanpach:
@bshaffer 5.4 and 5.5 are removed from the workflows and composer.json. All test pass now!

tests/Google/Task/RunnerTest.php
```diff
@@ -288,14 +288,15 @@ public function testCurlTimeouts($config, $minTime)
...
-    new Google_Task_Runner(
-        $this->retryConfig,
-        '',
-        array($this, 'testBadTaskConfig')
-    );
+      new Google_Task_Runner(
+          $this->retryConfig,
+          '',
+          array($this, 'testBadTaskConfig')
+      );
```
bshaffer, line 299: This whitespace shouldn't change

bshaffer (commented):
Just one issue with the whitespace and we can merge!

bshaffer (approved):

README.md
```diff
@@ -11,14 +11,14 @@ The Google API Client Library enables you to work with Google APIs such as Gmail
 
 These client libraries are officially supported by Google.  However, the libraries are considered complete and are in maintenance mode. This means that we will address critical bugs and security issues but will not add any new features.
 
-**NOTE** The actively maintained (v2) version of this client requires PHP 5.4 or above. If you require support for PHP 5.2 or 5.3, use the v1 branch.
+**NOTE** The actively maintained (v2) version of this client requires PHP 5.6 or above. If you require support for PHP 5.2 or 5.3, use the v1 branch.
```
tyteen4a03, line 14: I think you're missing a note for the last version that supports PHP 5.4 / 5.5

PR #2057, opened 2021-03-09
docs: update readme (how to use a specific JSON key)

kumamidori:
Update the README (how to use a specific JSON key)


README.md
```diff
@@ -217,6 +217,19 @@ calls return unexpected 401 or 403 errors.
     $client->setSubject($user_to_impersonate);
     ```
 
+#### How to use a specific JSON key
+
+If you want to a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, you can do this:
+
+```
```
bshaffer, line 224: set the syntax for php here:

```
```php
```

kumamidori: @bshaffer Fixed and git squashed my commit. Thanks.

kumamidori:
This option is useful. Some projects choose "Secrets as a service" solutions instead of environment variables.

- ref. [Best practices for managing & storing secrets like API keys and other credentials \[2020\]](https://blog.gitguardian.com/secrets-api-management/#avoid-git-add)

PR #2062, opened 2021-03-16
chore(docs): bump version in readme

bshaffer:


README.md
```diff
@@ -40,7 +40,7 @@ composer installed.
 Once composer is installed, execute the following command in your project root to install this library:
 
 ```sh
-composer require google/apiclient:"^2.7"
+composer require google/apiclient:"^2.9"
```
jdpedrie, line 43: Would `^2` be more forward-compatible with future minor versions?

jdpedrie (approved):

README.md
bshaffer: The point here is to make sure they wouldn't get a version _lower_ than 2.9

PR #2113, opened 2021-08-04
Fixing phpdoc for Resource::call method return type

amackiewicz:
psrFound that issue by the way of using Google_Service_Drive->files->export. 

Files::export calling Resource::call and returning its output. In normal circumstances it returns psr ResponseInterface, just in some cases (when request is defered) it returns RequestInterface. 

Current @return in Resource::call method has 2 issues:
1. Its solid class Response instead of ResponseInterface
2. Its cheating IDE which dont know that File::export could return ResponseInterface which end up with code as following. Of course its triggers `ERROR: TypeError: Return value of Foxstone\BackendPHP\service\GoogleDriveService::exportFileContents() must be an instance of GuzzleHttp\Psr7\Request, instance of GuzzleHttp\Psr7\Response returned`

Example of generated code (type returns by exportFileContents method was inherited from types returns from File::export (and then from Resource::call) which is wrong): 

```
    /**
     * @throws Exception
     */
    public function exportFileContents(
        string $fileId,
        string $exportMimetype
    ): ResponseInterface {
        $this->connect();

        return self::$GoogleServiceDrive->files->export(
            $fileId,
            $exportMimetype,
            ['alt' => 'media']
        );
    }
```

### DIFF
--- a/src/Service/Resource.php
+++ b/src/Service/Resource.php
@@ -80,7 +80,7 @@ class Resource
    * @param $name
    * @param $arguments
    * @param $expectedClass - optional, the expected class name
-   * @return Request|$expectedClass
+   * @return $expectedClass|ResponseInterface|RequestInterface|array
    * @throws \Google\Exception
    */
   public function call($name, $arguments, $expectedClass = null)
--- a/src/Client.php
+++ b/src/Client.php
@@ -856,7 +856,7 @@ class Client
    * @param $request RequestInterface|\Google\Http\Batch
    * @param string $expectedClass
    * @throws \Google\Exception
-   * @return object of the type of the expected class or Psr\Http\Message\ResponseInterface.
+   * @return $expectedClass|ResponseInterface|array
    */
   public function execute(RequestInterface $request, $expectedClass = null)
   {


amackiewicz:
@googlebot I signed it!

bshaffer:
@amackiewicz This looks great, although I'm not sure when it would return `array` for these functions. Can you elaborate?

amackiewicz:
Hey @bshaffer ! well, me too :) but problem was with REST::execute returning type in phpdoc (https://github.com/googleapis/google-api-php-client/blob/master/src/Http/REST.php#L44). 
I've changed phpdoc also for that method, check it out plz

src/Service/Resource.php
```diff
@@ -80,7 +80,7 @@ public function __construct($service, $serviceName, $resourceName, $resource)
    * @param $name
    * @param $arguments
    * @param $expectedClass - optional, the expected class name
-   * @return Request|$expectedClass
+   * @return $expectedClass|ResponseInterface|RequestInterface|array
```
bshaffer, line 83: remove `array` from this typehint also
```suggestion
   * @return mixed|$expectedClass|ResponseInterface|RequestInterface
```

amackiewicz: Hey @bshaffer , seems like you've fixed that already

amackiewicz:
any updates here @bshaffer  ? :)

PR #2115, opened 2021-08-11
Improve PHPUnit test suite

swissspidy:
This PR makes it easier to test the code base across different PHP versions by adopting PHPUnit Polyfills, which offers a number of polyfills for functionality which was introduced, split up or renamed in PHPUnit.

In other words, tests can be writing using assertions available in the latest PHPUnit version, even when running on older PHPUnit versions.

This removes the need for patching PHPUnit like was done before, and makes it easier to run tests against PHP releases using the most up-to-date versions of PHPUnit.

As proof, tests are now running against PHP 8.1 nightly (and they all pass!)

Some special logic was needed to make [Prophecy](https://github.com/phpspec/prophecy-phpunit) work across all supported PHP versions, but otherwise the changes should be rather straightforward.


bshaffer:
Hi! Thank you for your work on this, it's really helpful. However, the `8.1 --prefer-lowest` test is [throwing a fatal error](https://github.com/googleapis/google-api-php-client/pull/2115/checks?check_run_id=3299641890#step:7:4) because you've configured the 8.1 tests to ignore failures

swissspidy:
Ah, good catch. Need to use a more recent PHPUnit version there. Diesn‘t make sense to try running an okder one anyway. Will fix it 🙂

swissspidy:
> Hi! Thank you for your work on this, it's really helpful. However, the `8.1 --prefer-lowest` test is [throwing a fatal error](https://github.com/googleapis/google-api-php-client/pull/2115/checks?check_run_id=3299641890#step:7:4) because you've configured the 8.1 tests to ignore failures

Fixed now!

bshaffer (approved):

Issue #2099, opened 2021-07-06
Guzzlehttp/psr7 2.x has been released


brightjune:
Thanks for stopping by to let us know something could be better!

```
Using version ^2.10 for google/apiclient
./composer.json has been updated
Running composer update google/apiclient
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - google/apiclient[dev-master, v2.10.0, ..., v2.10.1] require guzzlehttp/psr7 ^1.2 -> found guzzlehttp/psr7[1.2.0, ..., 1.x-dev] but the package is fixed to 2.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.
    - google/apiclient 2.x-dev is an alias of google/apiclient dev-master and thus requires it to be installed too.
    - Root composer.json requires google/apiclient ^2.10 -> satisfiable by google/apiclient[v2.10.0, v2.10.1, 2.x-dev (alias of dev-master)].

Use the option --with-all-dependencies (-W) to allow upgrades, downgrades and removals for packages currently locked to specific versions.

Installation failed, reverting ./composer.json and ./composer.lock to their original content.
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

bshaffer:
This has long since been resolved

PR #2128, opened 2021-09-13
feature: allow usage of guzzlehttp/psr7 v2 

n00dl3:
replace deprecated function calls & bump min version of guzzlehttp/psr7 to  1.7+ or 2.0
related to #2099


n00dl3:
@googlebot I signed it!

n00dl3:
@googlebot I fixed it.

tests/Google/Service/ResourceTest.php
```diff
@@ -51,7 +51,7 @@ class TestMediaTypeStream extends Stream
 {
   public $toStringCalled = false;
 
-  public function __toString()
+  public function __toString():string
```
n00dl3, line 54: Oops, I didn't want to do that, I will remove this return type.

dwsupplee (approved):

dwsupplee:
Thanks @n00dl3, this is great 👍.

n00dl3:
Hello, I added a version check in the test, tests pass for 7.3+ I had difficulties compiling 7.2, but I think it should be okay. 
Unfortunately, I was not able to run the tests on php 5.6 due to some php7+ syntax in other test files, maybe I am misunderstanding something.

n00dl3:
This is the error I encountered, I guess tests are failing on php 5.6 since https://github.com/googleapis/google-api-php-client/commit/1486680805450d6fd3bf0492d6152fae8712b74a .

tests/Google/Service/ResourceTest.php
```diff
@@ -47,16 +47,36 @@ public function __construct(Client $client)
...
+    }
+// version is gte 7.2, so psr7 version can be:
+// * v2, so __toString() return is typed
+// * v1, so we can safely override __toString() to make it return string
+} else {
+    class TestMediaTypeStream extends Stream
+    {
+        public $toStringCalled = false;
+
+        public function __toString():string
```
dwsupplee, line 73: @n00dl3, it looks like this may be the culprit for the 5.6 test failure. Could we re-evaluate this approach?

n00dl3: Honestly I have no idea, without putting the class in an external file and require it conditionally like :
```php
// if php version is lower than 7.2, then psr7 v1 is installed,
// so __toString() return is not explicitly typed in parent class
if(version_compare(PHP_VERSION,'7.2.0','<')) {
    require_once __DIR__.'/TestMediaTypeStreamv5+.php';
// version is gte 7.2, so psr7 version can be:
// * v2, so __toString() return is typed
// * v1, so we can safely override __toString() to make it return string
} else {
    require_once __DIR__.'/TestMediaTypeStreamv7+.php';
}
```

Anyway, tests will fail right after that because of the `setUp():void` declaration.

Maybe I should just stick to the ` public function __toString():string` declaration compatible with php 7+ ?

@dwsupplee An opinion, please ?

bshaffer: Fixed in 43dd3443602e475951966da3be6745b73f49ec33

PR #2130, opened 2021-09-17
Warn about null properties that may actually have values, but were hydrated as null because those fields are not present in the Response.

edsonhoraciojunior:


edsonhoraciojunior:
@googlebot I signed it!

edsonhoraciojunior:
@googlebot I fixed it.

edsonhoraciojunior:
@googlebot I fixed it!

edsonCodelathe:
@googlebot I consent.

bshaffer:

I don't know how useful this paragraph is. Firstly, because the user would have to explicitly and knowingly supply the `fields` parameter, which isn't obvious by the wording here. Secondly because there's no example of what to do in the case mentioned.

bshaffer (approved):

PR #2134, opened 2021-09-30
fix: ensure refresh token remains after get token callback

marcoocram:
When a token expires and refresh token is provided, first time on "authorize" refresh the token. 

However, this dispatch the "token callback" to save the new token (and get the new expiration time) but it doesn't set again this refresh_token, so the token keeps without this data and on the next expiration, it will not refresh the token for this reason.

With this PR we ensure that the refresh token remains on the token property always.

EDIT: I've added merge old token configuration for scopes, grant type...

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -188,11 +188,14 @@ class Client
     if (is_null($this->config['token_callback'])) {
       $this->config['token_callback'] = function ($cacheKey, $newAccessToken) {
         $this->setAccessToken(
-            [
-              'access_token' => $newAccessToken,
-              'expires_in' => 3600, // Google default
-              'created' => time(),
-            ]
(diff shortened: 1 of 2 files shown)


bshaffer:
I am a bit nervous about this change. I think we could be a bit smarter about it... for instance...

```php
$accessToken = [
    'access_token' => $newAccessToken,
    'expires_in' => 3600, // Google default
    'created' => time(),
];
$previousAccessToken = $this->getAccessToken();
if (isset($previousAccessToken['refresh_token'])) {
    $accessToken['refresh_token'] = $previousAccessToken['refresh_token'];
}
$this->setAccessToken($accessToken);
```

marcoocram:
Two months....this resolution.

PR #2153, opened 2021-11-14
feat: allow credentials object

bshaffer:
Allows for a `Google\Auth\CredentialsLoader` instance to be passed into the constructor of `Google\Client` using the `credentials` option, which currently exists but only supports an array or string.

```php
$client = new Google\Client([
    'credentials' => new AnonymousCredentials()
]);
```

This also solves #2142, which provides a way to enable self-signed JWT, but in a more general (and useful) way. 

```php
$scope = 'https://www.googleapis.com/auth/cloud-platform';
$jsonKey = json_decode(file_get_contents('./path/to/key.json'), true);
$credentials = Google\Auth\ApplicationDefaultCredentials::makeCredentials([
    'scope' => $scope,
    'jsonKey' => $jsonKey,
]);

// Enable self-signed JWT
$credentials->useJwtAccessWithScope();

$service = new Google\Service\BigQuery([
    'credentials' => $credentials
]);
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -87,6 +87,11 @@ class Client
   private $logger;
 
   /**
+   * @var CredentialsLoader $credentials
+   */
+  private $credentials;
+
+  /**
    * @var boolean $deferExecution
    */
   private $deferExecution = false;
@@ -114,8 +119,9 @@ class Client
           'client_id' => '',
           'client_secret' => '',
 
-          // Path to JSON credentials or an array representing those credentials
-          // @see Google\Client::setAuthConfig
+          // Can be a path to JSON credentials or an array representing those
+          // credentials (@see Google\Client::setAuthConfig), or an instance of
+          // Google\Auth\CredentialsLoader.
           'credentials' => null,
           // @see Google\Client::setScopes
           'scopes' => null,
@@ -175,7 +181,11 @@ class Client
     );
 
     if (!is_null($this->config['credentials'])) {
-      $this->setAuthConfig($this->config['credentials']);
+      if ($this->config['credentials'] instanceof CredentialsLoader) {
+        $this->credentials = $this->config['credentials'];
+      } else {
+        $this->setAuthConfig($this->config['credentials']);
+      }
       unset($this->config['credentials']);
     }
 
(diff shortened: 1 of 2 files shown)


noahdietz (approved):

tmatsuo (commented):
With this change, new user code can pass a CredentialLoader.

What happens if you use the new user code with an old client library? My guess is the user will get a confusing, non actionable message (I may be wrong).

bshaffer:
> What happens if you use the new user code with an old client library? My guess is the user will get a confusing, non actionable message (I may be wrong).

Unfortunately, you're correct... a user would get an error like "Cannot use object Google\Auth\Credentials\ServiceAccountCredentials as array". 

I don't see this as a blocker for two reasons: 1. we have the options documented for what's accepted, and 2) in the user guide where we're using this feature (configuring Self-Signed JWTs), we explicitly state that they will need to upgrade to `v2.12` of this library to use the feature.

bshaffer:
> What happens if you use the new user code with an old client library? My guess is the user will get a confusing, non actionable message (I may be wrong).

Unfortunately, you're correct... a user would get an error like "Cannot use object Google\Auth\Credentials\ServiceAccountCredentials as array". 

I don't see this as a blocker for two reasons: 1. we have the options documented for what's accepted, and 2) in the user guide where we're using this feature (configuring Self-Signed JWTs), we explicitly state that they will need to upgrade to `v2.12` of this library to use the feature.

tmatsuo:
@bshaffer I was wondering if there's a better way for handling this change, but... maybe the benefit outweighs the negatives.

tmatsuo (approved):

bshaffer:
@tmatsuo we could certainly use a different option... something like `credentialsLoader`. That could potentially work better... the reason I went with this method is just that the term "credentials" is used to mean the `CredentialsLoader` object in our `google/cloud` library, so I was trying to stay consistent

realtebo:
what is the last version without this issue?

bshaffer:
@realtebo try [v2.12.1](https://github.com/googleapis/google-api-php-client/releases/tag/v2.12.1)

Issue #2155, opened 2021-11-16
Allow Users to Supply Additional Guzzle Client's Options


iastate-zuo:
At our organization, all server-initiated requests are routed through a reverse-proxy server, which means we need to set the `proxy` options for the HTTP client to properly send outbound HTTP requests. 

Unfortunately, the [`Google\Client`](https://github.com/googleapis/google-api-php-client/blob/master/src/Client.php) class does not support supplying additional Guzzle options when setting it up, and the usual workaround to supply a custom HTTP client to the `Google\Client` is out of the question because we do not have direct control of the `google-api-php-client` package -- it's been used as a dependency of a 3rd party package. Consequently, we can't create a Guzzle Client with reverse-proxy settings and hand it over to the `Google\Client` because there is more than one layer of abstractions between our server code and `google-api-php-client`.

A possible solution to this problem is to add a `guzzle_option` config item to the `$config` variable, such that when creating the default [HTTP client](https://github.com/googleapis/google-api-php-client/blob/master/src/Client.php), we can merge the guzzle options with the `$options`, and supply the updated options to the Guzzle client constructor. Since most of the 3rd party packages will pass through `$config` variables, our server code will be able to hand over the desired Guzzle settings to the `Google\Client`.

For example, I'm thinking of some code like this:

```
protected function createDefaultHttpClient()
{
    // ... code to initiatialize $opions 
    $options = array_merge($options, $this->config['guzzle_options']);
    return new GuzzleClient($options);
}
``` 

I will summarize the solution to a PR to better illustrate this idea.

bshaffer:
There is no need for this as an option, when you can set the client to a custom Guzzle Client at any time using `Google\Client::setHttpClient`:
```php
$client = new Google\Client();
$client->setHttpClient(new GuzzleHttp\Client($myGuzzleOptions));
```

PR #2156, opened 2021-11-16
Allow Users to Set Additional Guzzle Options

iastate-zuo:
This PR will summarize the `$config`-based solution I'm thinking of for issue #2155, where we need to pass the `proxy` settings to the default HTTP client (i.e., the Guzzle client) because our server code doesn't have direct control of the `Google\Client` class (i.e., we can't create a custom HTTP client with the proper proxy settings and hand it over to `Google\Client` in our use case), and we can only interact with this client class via `$config` settings.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -169,7 +169,10 @@ class Client
 
           // Setting api_format_v2 will return more detailed error messages
           // from certain APIs.
-          'api_format_v2' => false
+          'api_format_v2' => false,
+          
+          // Setting additional guzzle options for the default http client
+          'guzzle_options' => [],
         ],
         $config
     );
@@ -1186,6 +1189,11 @@ class Client
     } else {
       throw new LogicException('Could not find supported version of Guzzle.');
     }
+
+    $options = array_merge(
+      $options,
+      $this->config['guzzle_options'],
+    );
 
     return new GuzzleClient($options);
   }


iastate-zuo:
Just signed CLA, please relaunch the CLA scan again, thanks!

iastate-zuo:
@googlebot I signed it!

bshaffer:
There is no need for this as an option, when you can set the client to a custom Guzzle Client at any time using `Google\Client::setHttpClient`:
```php
$client = new Google\Client();
$client->setHttpClient(new GuzzleHttp\Client($myGuzzleOptions));
```

iastate-zuo:
I've seen this workaround somewhere else. However, this does not help our use case: our organization use the `laravel-analytics` package to query the analytics data, and the package does not expose the `$client` object to the package users. Consequently, we have no control or direct way of accessing the `$client` and providing a customized `GuzzleClient` to it. That's why we'd want a more direct way of passing `Guzzle` options when the HTTP client is created.

bshaffer:
I don't understand... Passing the options to this class (what you suggest), or creating a NEW client and passing the new client to this class (what i suggest and what is currently supported) is functionally identical.

iastate-zuo:
Yes, functionality-wise, the two options are identical. The problem is that there are legit use cases where the consumer has access to only one of the two options (i.e., via an option setting).

Let's consider an example -- say, there is a popular 3rd party package, which uses the `Google API Client` with the below implementation:

```
use Google\Client as GoogleClient;

namespace ThirdPartyPackage;

class ThirPartyClient
{
    protected $client;

    public function __construct(array $config)
    {
        $this->client = new GoogleClient($config['client_config_options']);
    }

    // ... and $client is never publicly exposed
}
```

And in our application code, we will create the `ThirdPartyClient` in our controllers. However, since `$client` is protected and hence not accessible, our application is not able to use the `setHttpClient` API from the `$client` object:

```
use ThirdPartyPackage\ThirdPartClient;

namespace App

class Controller
{
    protected $service;

    public function __construct()
    {
        $config = config('thirdPartyConfig');
        $this->service = new ThirdPartClient($config);

        // error: $client is a protected attribute, and it can't be accessed
        $this->service->client->setHttpClient(new GuzzleHttp\Client($config['client_config_options']));
    }
}
```
Our best bet for this use case is to let the default HTTP client be created with the options passed down in the dependency cascade without altering all the classes/packages to expose the `Google\Client` object along the way.

Issue #2162, opened 2021-12-01
Error: Call to undefined method Google\Client::isUsingJwtWithScope()


stephanvierkant:
#2153 created an issue because `\Google\Client` doesn't have a method called `isUsingJwtWithScope`:

> Error: Call to undefined method Google\Client::isUsingJwtWithScope()

https://github.com/googleapis/google-api-php-client/blob/8865cab2efbd63b40f88552886865f31bd7f3edd/src/Client.php#L1258

Ping @bshaffer 

Maybe some kind of static analysis should be added to the test suite?

realtebo:
[2021-12-02 00:59:21] production.ERROR: Call to undefined method Google\Client::isUsingJwtWithScope() {"userId":"d25d0506-8a96-3675-99b1-92dd4843dc06","exception":"[object] (Error(code: 0): Call to undefined method Google\\Client::isUsingJwtWithScope() at /vendor/google/apiclient/src/Client.php:1258)
[stacktrace]
#0 /vendor/google/apiclient/src/Client.php(436): Google\\Client->createApplicationDefaultCredentials()
#1 /vendor/google/apiclient/src/Client.php(911): Google\\Client->authorize()


Same problem here.
Don't you run at least a basic test suite before release ?! You broke our code

jdevinemt:
I also have this issue. I'm also seeing `#[\ReturnTypeWillChange]` sprinkled throughout the code, which appears to be specific to php 8.1+. But in the package description it should work with older php versions.

sheyDev:
Call to undefined method Google\Client::isUsingJwtWithScope();

Running into the same problem. Any fix for this?

jdevinemt:
> 

You can just revert to v2.11.0 until it's fixed.

bshaffer:
Fixed in [v2.12.1](https://github.com/googleapis/google-api-php-client/releases/tag/v2.12.1)

stephanvierkant:
> Fixed in [v2.12.1](https://github.com/googleapis/google-api-php-client/releases/tag/v2.12.1)

Thanks for fixing this quickly!

Will there be any measures taken to prevent this kind of bugs? Even the most simple kind of static analysis would have found this.

bshaffer:
I agree, and am also surprised it slipped through, since we have an extensive testing suite... there must be something unexpected happening with regards to this error, unfortunately.

stephanvierkant:
Can I help with implementing static analysis, like phpstan?

[...]

PR #2166, opened 2021-12-03
Added PHPStan

stephanvierkant:
See #2162 for context.

What I've done so far:
- Added PHPStan with a baseline and already fixes some low hanging fruit. 
- Removed support for EOL-versions of PHP. That includes PHP 7.3, that will become EOL in the very near future.
- That also means I had to drop support for Guzzle5, which is EOL for a long time as well.

I'm still struggling with one issue:

> TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given

I'll pick this up later (suggestions are welcome!) but since this PR ended up more complex than I hoped, I wanted to share this progress with you. Please let me know what you think!


mfn:
👎🏼 on the PR, not on the idea.

It does too much, it's too much burden on the reviewer. Decisions to drop support for certain PHP versions should be independent, etc.

Ideally a PR would be made doing on thing and doing and explaining it well, and not so many things.

Just my 2c.

stephanvierkant:
I understand. If one agrees on dropping support for these old PHP versions (including Guzzle5) and using PHPStan as static analyzer, I can split up this PR.

bshaffer:
As mentioned in the other PR, we will be continuing to support PHP 7.3 for the foreseeable future. I am okay dropping Guzzle5 and PHP 5.6, but we don't want to be aggressive with our PHP version deprecation because this is a client library and we want to support as many users as possible.

PR #2167, opened 2021-12-04
Remove support EOL-versions of PHP (and Guzzle5)

stephanvierkant:
On 6 Dec 2021, PHP 7.3 becomes End of Life (EOL). That means only 7.4, 8.0 and 8.1 are supported and will get security updates. So let's end support for other versions in this library as well.

This PR removes support for PHP 5.6, 7.0, 7.1, 7.2, 7.3. It also removes support for Guzzle5, because it's EOL for a long time as well and only works on PHP 5.x.

After this PR is merged, I will continue working on adding PHPStan to this library: #2166.


bshaffer:
The unfortunate reality is there are most likely many people using this library on EOL'd versions of PHP, and we want to support them. We are not ready to drop support up to 7.3, as we do not want to aggressively deprecate support for PHP for this library.

Dropping support for PHP 5, and therefore Guzzle5, is a potential.

stephanvierkant:
Sorry, I'm not spending time on supporting this very very old versions of PHP, so I closed my other PR as well.

People should not use PHP versions that are EOL for such a long time.

PR #2173, opened 2021-12-07
Fix `src/aliases.php`

WinterSilence:
1. added class autoloading for `class_exists()`
2. removed class preloading by `class_alias()` 
3. added removing variables after work

### DIFF
--- a/src/aliases.php
+++ b/src/aliases.php
@@ -7,7 +7,8 @@ if (!class_exists('Google_Client')) {
             new RecursiveDirectoryIterator(
                 __DIR__,
                 RecursiveDirectoryIterator::UNIX_PATHS
-            )
+            ),
+            RecursiveIteratorIterator::LEAVES_ONLY
         ),
         '/((?:[A-Z][a-zA-Z\d]+\/)*[A-Z][a-zA-Z\d]+)\.php$/',
         RegexIterator::GET_MATCH,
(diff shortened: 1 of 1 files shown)


bshaffer:
Hi @WinterSilence ! Thank you for your contribution

I am a little unsure what is being fixed here. I think I may be missing something, so maybe there's a use-case I don't see where you're seeing an error? 
1. Ensuring `class_exists` doesn't execute the autoloader (second argument `false`) is intentional because it is used in the case of `preloading`. There is no reason to look in the other autoloaders for this class, because it will never exist unless it's preloaded.
2. I see how using recursion / PSR-0 might be considered better in case new classes are added and also because the existing classes can be autoloaded instead of creating the alias up front, but new classes will not need to maintain Backwards Compatibility, and we only have 19 aliases being added, so I believe that a classmap is faster and more efficient
2. The check to see if `Google_Composer_Task` exists is unnecessary because it will never exist, and the check if `Google\Composer\Task`, likewise, will always exist. So how is that making it only autoload in "development mode"?

Unsetting the variables seems like an improvement, but the other changes I am unsure how they will improve things.

WinterSilence:
@bshaffer 1-2. I thought this was done to support legacy code & code style
3. I missed changes in `composer.json` - composer task must be excluded in production

PR #2182, opened 2021-12-22
Deleting used variables and unusable code in `aliases.php`

WinterSilence:


src/aliases.php
```diff
@@ -40,26 +41,3 @@ class_alias($class, $alias);
...
-  class Google_Http_REST extends \Google\Http\REST {}
-  class Google_Model extends \Google\Model {}
-  class Google_Service extends \Google\Service {}
-  class Google_Service_Exception extends \Google\Service\Exception {}
-  class Google_Service_Resource extends \Google\Service\Resource {}
-  class Google_Task_Exception extends \Google\Task\Exception {}
-  interface Google_Task_Retryable extends \Google\Task\Retryable {}
-  class Google_Task_Runner extends \Google\Task\Runner {}
-  class Google_Utils_UriTemplate extends \Google\Utils\UriTemplate {}
-}
```
bshaffer, line 65: This is here for IDEs, without it they don't know the aliases exist

bshaffer (changes requested):

src/aliases.php
WinterSilence: @bshaffer strange solution.. why not stubs?

WinterSilence: @bshaffer also it's not correct because we don't have that's classes, it's must be something like `use Google\AccessToken\Revoke as Google_AccessToken_Revoke;`

bshaffer: These classes are added as aliases in the code above this. That's why `aliases.php` exists, to create the aliases that are defined in this list. See #1995

WinterSilence: @bshaffer it's not same:
~~~php
class A {}
class_alias(A::class, 'B');
var_dump(get_class(new B()) === B::class); // false
class C extends A {}
var_dump(get_class(new C()) === C::class); // true
~~~

WinterSilence: @bshaffer we can generate stubs for IDE, see https://psalm.dev/docs/running_psalm/plugins/authoring_plugins/

bshaffer: I do not understand what you're trying to achieve or what you think the problem is with the current implementation. I'd prefer to keep it as is unless there is a specific problem you're trying to fix.

WinterSilence: @bshaffer I'm removed redundant code, for ide need generate stubs

bshaffer: These lines are not redundant. They are here so IDEs know the aliases exist.

I don't know what generating stubs would accomplish that is not already accomplished here by the lines of code you removed.

PR #2183, opened 2021-12-22
Remove redundand and invalid "classmap" in `composer.json`

WinterSilence:
"src/aliases.php" is not classmap

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,9 +36,6 @@
         },
         "files": [
             "src/aliases.php"
-        ],
-        "classmap": [
-            "src/aliases.php"
         ]
     },
     "extra": {


bshaffer:
The tests fail without this line. I'm not sure why, but I know it is necessary for these classes to function as expected. I believe its because they need to be called when the autoloader is invoked.

WinterSilence:
@bshaffer reason failed tests:
> 1) Google\Tests\Task\ComposerTest::testE2EBCTaskName
Failed asserting that file "/tmp/test-1912002904/vendor/google/apiclient-services/src/YouTube.php" does not exist.

has nothing to do with this PR

bshaffer:
@WinterSilence yes it does, it is happening as a direct result of your PR.

WinterSilence:
@bshaffer :facepalm it's happened because you miss
 `"autoload-dev": {
        "psr-4": {
            "Google\\Tests\\": "tests/Google/"
        },
        "classmap": ["tests/BaseTest.php"]
},
`

PR #2184, opened 2021-12-22
Optimize `Model::camelCase()`

WinterSilence:


bshaffer:
I appreciate this PR, but since this library has been around a very long time, Id rather not risk changing something this fundamental unless the benefits are very clear. For this change, I do not think the optimizations would be significant enough to justify the risk.

WinterSilence:
@bshaffer what risks are you talking about? tests work fine

bshaffer:
Yes but as we both know, it's possible for tests to miss things. Thanks again for your contribution.

WinterSilence:
@bshaffer based on this logic, any change is impossible. it's wrong way, right way - extend test to covering more cases.

PR #2185, opened 2021-12-23
Fix tests.yml

WinterSilence:
Fix for `matrix.composer-flags == '--prefer-lowest'` always returned false


bshaffer:
Thanks for submitting this! More cleanup is required to this file than is done here in this PR, so I've created a new PR in #2190 to address this.

PR #2191, opened 2022-01-04
micro optimisation of loop array

brytonl33t:


mfn:
> micro optimisation

This is not just an optimization, but also change of behaviour

PR #2218, opened 2022-02-24
Check token has expires_in attribute

dmitryuk:
Related https://github.com/googleapis/google-api-php-client/issues/2217

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -519,7 +519,7 @@ class Client
     if ($token == null) {
       throw new InvalidArgumentException('invalid json token');
     }
-    if (!isset($token['access_token'])) {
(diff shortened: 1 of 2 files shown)


dmitryuk:
CLA signed

bshaffer:
This check may affect code paths that were previously unaffected by this error, therefore causing a backwards-compatibility issue. I suggest adding an `isset` before line 574 instead:

```php
if (!isset($this->token['expires_in'])) {
    // if the token does not have an "expires_in", then it's considered expired
    return true;
}

return ($created + ($this->token['expires_in'] - 30)) < time();
```

dmitryuk:
@bshaffer, agree with you. Added new changes to PR

bshaffer (approved):

PR #2244, opened 2022-04-14
Fix #2243

### LINKED ISSUE #2243: Invalid argument supplied for foreach() in firebase/php-jwt/src/JWT.php:422
Pull request https://github.com/googleapis/google-api-php-client/pull/2235 updated this package's `composer.json` file to allow `firebase/php-jwt` v6.

`firebase/php-jwt` v6 changed the method signatures for both `JWT::encode()` and `JWT::decode()`. It seems that the calls in this library to this method were not updated accordingly.

From the [firebase/php-jwt change log](https://github.com/firebase/php-jwt/releases/tag/v6.0.0)

> The second argument of JWT::decode now must be Firebase\JWT\Key or array<string, Firebase\JWT\Key> (see https://github.com/firebase/php-jwt/pull/376)
> JWT::encode requires third argument $alg (see https://github.com/firebase/php-jwt/pull/376)

The offending decode call is in `src/AccessToken/Verify.php(106): Firebase\\JWT\\JWT::decode()`

I will follow up this issue with a PR.

#### Environment details

  - OS: Ubuntu 20.04
  - PHP version: 7.4.28
  - Package name and version: 2.12.2

bstanleynep:
This PR is a follow up to issue #2243. It looks like, at the very least, v5 and v6 of firebase/php-jwt accept an instance of `Firebase\JWT\Key` as the 2nd argument to `JWT::decode()`. All of the tests pass but at least one of them should have been failing before this change if you have firebase/php-jwt v6 installed. 

I can add/update tests if required.

### DIFF
--- a/src/AccessToken/Verify.php
+++ b/src/AccessToken/Verify.php
@@ -19,6 +19,7 @@
 namespace Google\AccessToken;
 
 use Firebase\JWT\ExpiredException as ExpiredExceptionV3;
+use Firebase\JWT\Key;
 use Firebase\JWT\SignatureInvalidException;
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
@@ -100,9 +101,8 @@ class Verify
     foreach ($certs as $cert) {
       try {
         $payload = $this->jwt->decode(
-            $idToken,
-            $this->getPublicKey($cert),
-            array('RS256')
+          $idToken,
+          new Key($this->getPublicKey($cert), 'RS256')
         );
 
         if (property_exists($payload, 'aud')) {


bstanleynep:
I did complete the CLA document about 2 minutes before I created this PR. Let me know if I need to do anything else.

bshaffer:
@bstanleynep good catch! this is a really important fix, so I don't want to wait too long to merge it, but if you _could_ add a test, that would be awesome!

bstanleynep:
@bshaffer no problem, I'll work on this today

bshaffer:
This is a bit more complicated than I originally thought (because we have to support older versions) So I've submitted a separate fix in #2245 

Thank you for pointing this issue out and cintributing a fix.

bshaffer:
Fixed in [v2.12.3](https://github.com/googleapis/google-api-php-client/releases/tag/v2.12.3)

PR #2245, opened 2022-04-18
fix: bad firebase call

bshaffer:
fix #2244 and #2243

We already have tests to cover this, but they were being skipped in CI because they require specific auth credentials


bshaffer:
Released in [v2.12.3](https://github.com/googleapis/google-api-php-client/releases/tag/v2.12.3)

bstanleynep:
@bshaffer ah, I see, that is a better fix for sure. Thanks for knocking this out!

PR #2251, opened 2022-04-21
chore: make whitespace consistent with our other libraries

bshaffer:
Finally make the whitespace consistent in this library with all our other libraries. This means moving all tabs to four spaces. What took us so long?


mfn:
HOLY DIFF!

But luckily there's [`?w=1`](https://github.com/googleapis/google-api-php-client/pull/2251/files?diff=unified&w=1)

bshaffer:
That's super helpful! Thanks @mfn

PR #2254, opened 2022-04-22
fix: add missing public var used by subclasses

bshaffer:
The classes generated in `google/apiclient-services` are setting a local property [$serviceName](https://github.com/googleapis/google-api-php-client-services/blob/main/src/Storage.php#L77) which isn't actually defined anywhere.


saranshdhingra (approved):
LGTM

PR #2271, opened 2022-06-02
Update return type of `Exception::getErrors()` to its correct structure

mrtus:
After the latest update static analysis tools hinted on this wrongly typed structure.

The returned structure of `Exception::getErrors()` is an array with a map of keys and values.

The change was introduced here https://github.com/googleapis/google-api-php-client/pull/2252
```
[
    {
        "domain": "global",
        "reason": "authError",
        "message": "Invalid Credentials",
        "locationType": "header",
        "location": "Authorization"
    }
]
```

### DIFF
--- a/src/Service/Exception.php
+++ b/src/Service/Exception.php
@@ -33,7 +33,7 @@ class Exception extends GoogleException
      * @param string $message
      * @param int $code
      * @param Exception|null $previous
-     * @param array<string,string> $errors List of errors returned in an HTTP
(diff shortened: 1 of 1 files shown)


mrtus:
Hi @bshaffer, What is necessary to get this change through? Am I missing some steps that would complete this change request for it to be verified & merged?

bshaffer (approved):

marekdedic:
Hi,
thank you for your work, this PR solved some "mystery" issues :) Could we please get this released?

Thanks!

PR #2275, opened 2022-06-08
Update accounts.google.com authorization URI

niallkennedy:
Update Google authorization URI to match `authorization_endpoint` value provided in the [accounts.google.com OIDC Discovery doc](https://accounts.google.com/.well-known/openid-configuration).


bshaffer (approved):

ziemkowski:
In case anybody stumbles upon this PR while researching an issue with some auth requests no longer returning a `refresh_token`, be aware that this change breaks old code that uses `setApprovalPrompt('force')` instead of the newer `setPrompt('consent')` method, as this endpoint no longer supports the approval prompt format. Changing your code to use the newer method should fix the issue.

PR #2288, opened 2022-06-29
Send no content-type header if there is no body

### LINKED ISSUE #2287: No-body request to YT api fails with supplied content-type header
I am trying to download YT captions for a given video. 

```php
$client = // client is configured with service account credentials or oauth2 token.
$response = (new Google_Service_YouTube($client))->captions->download($caption['id'])->getBody();
```

However this throws exception and YT response looks like this:
```
{
  "error": {
    "code": 400,
    "message": "Unsupported Output Format",
    "errors": [
      {
        "message": "Unsupported Output Format",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```
I traced issue down to this line https://github.com/googleapis/google-api-php-client/blob/1530583a711f4414407112c4068464bcbace1c71/src/Service/Resource.php#L206 
Apparently, setting `content-type` header makes YT api break.

I'd suggest adding this header conditionally as it seems like [that was the case](https://github.com/googleapis/google-api-php-client/commit/6dc8190553988377554d54048bd086b89c57a37b#diff-c42742883e8032aa0b5735c831478483a43d7ad2a2a8e3fa32b188ca4ec579d8L203) before upgrading to Guzzle6

kkopachev:
Fixes #2287

### DIFF
--- a/src/Service/Resource.php
+++ b/src/Service/Resource.php
@@ -203,7 +203,7 @@ class Resource
         $request = new Request(
             $method['httpMethod'],
             $url,
-            ['content-type' => 'application/json'],
+            $postBody ? ['content-type' => 'application/json'] : [],
             $postBody ? json_encode($postBody) : ''
         );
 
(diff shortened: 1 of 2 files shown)


bshaffer (approved):
Thanks for tracking this down!

PR #2290, opened 2022-07-05
Add http options to allow timeout

bernardwiesner:
As mentioned here: https://github.com/googleapis/google-api-php-client/issues/1767

There is no simple way to set a `timeout` or any other options to the guzzle client. 

The suggestion from @jdpedrie in above issue is not optimal since you lose the default options set by this package. With this change you can easily do:

```php
$client = new Google_Client([
  'http_options' => [
     'timeout' => 5
  ]
]);
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -115,6 +115,7 @@ class Client
             // Don't change these unless you're working against a special development
             // or testing environment.
             'base_path' => self::API_BASE_PATH,
+            'http_options' => null,
 
             // https://developers.google.com/console
             'client_id' => '',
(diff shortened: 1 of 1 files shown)


NSROHILLA82:
$Http invalid format $ PCr not valid with combined .jdk files.

bshaffer:
Hello @bernardwiesner!

So, we include a method to set the http client on the google client object:

```php
$http = new GuzzleHttp\Client([
    'timeout' => 5
]);
$client->setHttpClient($http);
```

This should be sufficient for what you're trying to do.

PR #2309, opened 2022-08-22
Fix unexpected null value in Service\Exception::$errors

tscni:
Based on the definition it should always be an array, but that's not reflected in `REST::getResponseErrors()`.

### DIFF
--- a/src/Service/Exception.php
+++ b/src/Service/Exception.php
@@ -33,8 +33,8 @@ class Exception extends GoogleException
      * @param string $message
      * @param int $code
      * @param Exception|null $previous
-     * @param array<array<string,string>> $errors List of errors returned in an HTTP
-     * response.  Defaults to [].
+     * @param array<array<string,string>>|null $errors List of errors returned in an HTTP
+     * response or null.  Defaults to [].
      */
     public function __construct(
         $message,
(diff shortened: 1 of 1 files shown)


bshaffer:
This is technically a breaking change, so it would be better to update the PHPdoc than to update the value here.

tscni:
> This is technically a breaking change, so it would be better to update the PHPdoc than to update the value here.

Wouldn't both variants be a breaking change?
In one case users rely on `Service\Exception:$errors` being an `array`, in the other users encountered an issue with it (like using it in a foreach in PHP 8) and know that it might be `null` for certain cases. (effectively type-widening vs type-narrowing)

Breaking the former expectation should affect a much higher percentage of the user base. Especially considering how unlikely it is that someone actually depends on the `null` behavior in any other way than just handling it as an empty array.

All depends on how you see the promise of your API of course.

bshaffer:
@tscni changing PHPDoc is not considered a breaking change, as changing it would not result in any _existing_ code based on our public API breaking. Changing the actual return type, however, could potentially break existing code. I hope this helps.

tscni:
Very well. I've adjusted the PR accordingly

bshaffer (approved):

bshaffer:
Thank you!

PR #2315, opened 2022-09-15
fix (style): lint errors

vishwarajanand:
fixing lint errors due to non closure of PHP tags

Fixes: https://github.com/googleapis/google-api-php-client/issues/2316

I ran `php -S localhost:8080 -t examples/` and the example pages load correctly on browser, without any errors on server.

Screenshot 2022-09-15 at 14 56 37

Screenshot 2022-09-15 at 14 56 53


saranshdhingra (approved):
LGTM

PR #2327, opened 2022-10-07
chore: change the prefix to `https://`

a1383n:
Change the link prefix of the [`homepage`](https://github.com/googleapis/google-api-php-client/blob/88cc63c38b0cf88629f66fdf8ba6006f6c6d5a2c/composer.json#L6) property in the `composer.json` file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "type": "library",
     "description": "Client library for Google APIs",
     "keywords": ["google"],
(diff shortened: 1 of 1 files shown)


vishwarajanand:
@a1383n thanks for making this contribution. Since we typically have `http://` in composer's homepage feld, I am closing thie PR.

Issue #2361, opened 2022-12-30
New authorization URI is a breaking change


bshaffer:
See https://github.com/googleapis/google-api-php-client/pull/2275

Updating to the v2 auth URL broke the behavior of using `setApprovalPrompt('force')`, which was deprecated in favor of instead of the newer `setPrompt('consent')`. 

In order to mitigate this breaking change, we could update the `setApprovalPrompt` function to call `setPrompt('consent')` instead.

vishwarajanand:
Reduced priority since it doesn't block release and is only applicable to a scenario with force prompt.

bshaffer:
As i haven't heard anyone complain or comment on this bug or it's fix, i am thinking we can just keep the current behavior. But we should update the phpdoc at the very least

PR #2362, opened 2022-12-30
fix: use prompt=consent for approval_prompt=force

bshaffer:
addresses #2361


vishwarajanand:
Seems only applicable to 3-legged OAuth, if this doesn't break anyone maybe we can close this PR.

PR #2430, opened 2023-05-01
chore: docs for accessing granted scope

bshaffer:
Document support for accessing granted scopes, which was added to `google/auth` in https://github.com/googleapis/google-auth-library-php/pull/441


bshaffer:
done in https://github.com/googleapis/google-api-php-client/pull/2432

PR #2438, opened 2023-05-10
feat: add pkce support and upgrade examples

bshaffer:
depends on
 - https://github.com/googleapis/google-auth-library-php/pull/454
 - https://github.com/googleapis/google-api-php-client/pull/2431

See b/270198724

cc @TimurSadykov @clundin25


src/Client.php
```diff
@@ -242,7 +242,7 @@ public function authenticate($code)
      * @param string $code code from accounts.google.com
      * @return array access token
      */
-    public function fetchAccessTokenWithAuthCode($code)
+    public function fetchAccessTokenWithAuthCode($code, $codeVerifier = null)
```
clundin25, line 245: Should the codeVerifier parameter be added to the doc comment above?

clundin25 (approved):

src/Client.php
bshaffer: Good catch, done!

TimurSadykov (approved):
LGTM

PR #2498, opened 2023-09-08
feat: add audience parameter to Client::verifyIdToken

### LINKED ISSUE #1900: Login by google works only when not using a client_id
I'm building a login by google for a mobile app and I have this weird issue that took from me 2 days to debug!! 

That's the method `verifyIdToken($id_token)` always returning false and the token is valid ( that comes from my mobile app ) as I have tested with google apies oAuth2

```dart
Dio dio = new Dio();
 Response response = await dio.get('https://www.googleapis.com/oauth2/v1/tokeninfo?id_token='+googleKey.idToken);
 print(response.data); //contains the token info
```

By Luck, I removed the client_id and lifted as empty str and walaa it works returning a user response payload 

I can't understand why is that!! I'm using the same client_id on a node app in production and it's working fine ( access token not id token ), and I'm moving from node to PHP so I used the same client_id to test.

The Code: 

```php
$params = $request->get_params();
$token  = isset( $params['token'] ) ? $params['token'] : false;

if ( ! $token ) {
	return new \WP_Error( 'no-token', __( 'No token received from Google', 'vivant' ) );
}

if ( ! class_exists( 'Google_Client' ) ) {
	include_once plugin_dir_path( HEADLESS_FILE ) . '/vendor/google/apiclient/src/Google/Client.php'; // change path as needed
}

$client = new \Google_Client();  // works without a client_id and not working with a client_id

// $clinet->setClientSer
try {
	$payload = $client->verifyIdToken( $token );

	return $payload; // for debugging the value
	if ( $payload ) {
		// whatever
	} else {
		return new \WP_Error( 'invalid-token', __( 'Token is not valid', 'vivant' ) );
	}
} catch ( \Exception $e ) {
	return new \WP_Error( $e->getCode(), $e->getMessage() );
}
```

bshaffer:
fixes https://github.com/googleapis/google-api-php-client/issues/1900 by allowing an `$audience` parameter that will be used instead of the Client ID configured in the client object.

### DIFF
--- a/src/Http/REST.php
+++ b/src/Http/REST.php
@@ -79,7 +79,6 @@ class REST
      */
     public static function doExecute(ClientInterface $client, RequestInterface $request, $expectedClass = null)
     {
-        // var_dump($request);exit;
         try {
             $httpHandler = HttpHandlerFactory::build($client);
             $response = $httpHandler($request);
--- a/src/Http/REST.php
+++ b/src/Http/REST.php
@@ -79,6 +79,7 @@ class REST
      */
     public static function doExecute(ClientInterface $client, RequestInterface $request, $expectedClass = null)
     {
+        // var_dump($request);exit;
         try {
             $httpHandler = HttpHandlerFactory::build($client);
             $response = $httpHandler($request);
--- a/src/AccessToken/Verify.php
+++ b/src/AccessToken/Verify.php
@@ -91,7 +91,7 @@ class Verify
      * accepted.  By default, the id token must have been issued to this OAuth2 client.
      *
      * @param string $idToken the ID token in JWT format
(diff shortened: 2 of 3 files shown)


src/Http/REST.php
```diff
@@ -79,6 +79,7 @@ public static function execute(
      */
     public static function doExecute(ClientInterface $client, RequestInterface $request, $expectedClass = null)
     {
+        // var_dump($request);exit;
```
hex7c0, line 82: Typo I guess 👼

bshaffer:
```suggestion
```

bshaffer:
Closing because there has been no activity. If this is something that would be helpful please open an issue on this repo with a feature request

Issue #2535, opened 2023-12-21
guzzle 6.x version requirement should be updated to avoid security vulnerabilities


gravelld:
`composer.json` lists:

        "guzzlehttp/guzzle": "~6.5||~7.0",

However, [this vulnerability](https://github.com/advisories/GHSA-25mq-v84q-4j7r) mandates >=6.5.8. So shouldn't it be:

        "guzzlehttp/guzzle": "~6.5.8||~7.0",

?

Apologies if I'm missing something...

PR #2536, opened 2023-12-21
fix: disallow vulnerable guzzle versions

bshaffer:
addresses #2535

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "firebase/php-jwt": "~6.0",
         "monolog/monolog": "^2.9||^3.0",
         "phpseclib/phpseclib": "^3.0.19",
-        "guzzlehttp/guzzle": "~6.5||~7.0",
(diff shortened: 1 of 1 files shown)


vishwarajanand (approved):

vojtasvoboda:
This change downgraded my `guzzlehttp/guzzle` library from 7.8.1 to 7.4.5. Why? Guzzle versions after the 7.4.5 shouldn't be affected by [CVE-2022-31090](https://github.com/advisories/GHSA-25mq-v84q-4j7r).

gravelld:
As per https://getcomposer.org/doc/articles/versions.md#tilde-version-range- the caret modifier should probably be used. Tilde in this case means `>=7.4.5 < 7.5`.

`^7.4.5` would mean `>=7.4.5 < 8.0`.

vojtasvoboda:
So I created PR #2546 with caret used.

PR #2550, opened 2024-01-08
fix: Add AllowDynamicProperties annotation in Service class

ajupazhamayil:
Resolves https://github.com/googleapis/google-api-php-client-services/issues/4106

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,6 +49,7 @@ use UnexpectedValueException;
  * The Google API Client
  * https://github.com/google/google-api-php-client
  */
+#[\AllowDynamicProperties]
 class Client
 {
     const LIBVER = "2.12.6";
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,7 +49,6 @@ use UnexpectedValueException;
  * The Google API Client
  * https://github.com/google/google-api-php-client
  */
-#[\AllowDynamicProperties]
 class Client
 {
     const LIBVER = "2.12.6";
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -49,6 +49,7 @@
  * The Google API Client
  * https://github.com/google/google-api-php-client
  */
+#[\AllowDynamicProperties]
```
bshaffer, line 52: Why are we adding dynamic properties to this class? This class is not extended by anything, nor does it extend anything. So any dynamic properties it has should be defined.

Are you seeing dynamic property warnings anywhere for this class?

bshaffer (changes requested):

src/Client.php
ajupazhamayil: You are right, removing this change.

bshaffer (commented):
Just to be clear before we merge this - where is it that dynamic properties are being used for `Service` classes? Can you identify where it's happening (so we can also confirm that this is the right action to take)?

ajupazhamayil:
> Just to be clear before we merge this - where is it that dynamic properties are being used for `Service` classes? Can you identify where it's happening (so we can also confirm that this is the right action to take)?

Hmm.. I was thinking about the `ShoppingContent` or similar classes extending the `Service` class. But, I just looked at them and the classes I should be concentrating are classes which extends `Model` class. Further, Model class already has the required annotation. Seems like this PR will not help in our scenario! My bad!

PR #2560, opened 2024-02-14
fix: batch requesting RequestInterface

gbretas:
This error causing the batch add requesting an  `Psr\Http\Message\RequestInterface;`. But this is an error.

### DIFF
--- a/src/Http/Batch.php
+++ b/src/Http/Batch.php
@@ -66,7 +66,7 @@ class Batch
         $this->batchPath = $batchPath ?: self::BATCH_PATH;
     }
 
-    public function add(RequestInterface $request, $key = false)
+    public function add($request, $key = false)
     {
         if (false == $key) {
             $key = mt_rand();


devstack-be:
I have the same problem and I have not yet managed to use the batch on the google calendar api. 

Google\Http\Batch::add(): Argument #1 ($request) must be of type Psr\Http\Message\RequestInterface, Google\Service\Calendar\Event given, called in...

devstack-be:
I fix my problem. @gbretas 
You need to add thoses lines:

```
            $this->httpClient->setDefer(true);
            $calendarService = new CalendarService($this->httpClient);
            $calendarService->getClient()->setUseBatch(true);
            $batch = $calendarService->createBatch();

...
                $request = $calendarService->events->insert(YOUR_CALENDAR_ID, $event);
                $requestId = YOUR_REQUEST_ID
                $batch->add($request, $requestId);
            try {
                $results = $batch->execute();
            }
            catch (Exception $e ){
                Log::info($e->getMessage());
            }
```

bshaffer:
@gbretas see @devstack-be's comment above, or see the batch example in `examples/batch.php`:

https://github.com/googleapis/google-api-php-client/blob/017400f609c1fb71ab5ad824c50eabd4c3eaf779/examples/batch.php#L54-L78

The proper usage is to pass the `RequestInterface` object after calling the method in batch-mode. The typehint we have there is correct.

PR #2562, opened 2024-02-20
chore: drop support for PHP 7.4

bshaffer:


Hectorhammett (approved):

Hectorhammett:
The 7.X tests are still running, Should we remove them? What is the protocol here @bshaffer

vishwarajanand:
README also needs to be changed: 
https://github.com/googleapis/google-api-php-client/blob/7e79f3d7be4811f760e19cc4a2c558e04196ec1d/README.md?plain=1#L28

vishwarajanand:
@Hectorhammett that is happening because php7.4 is mentioned in the repo settings. It can be fixed by removing it, but only admin can see/do this change.

vishwarajanand (approved):
Approving to unblock

PR #2587, opened 2024-04-19
docs: fix namespace slash problem

xialeistudio:
Hi Google Api PHP Client Team,

I found some problems with README.md. When I tried to integrate Oauth, I copied the code of README.md and I found an error that the class did not exist. I found out that it was because there was a missing slash. According to the PHP specification, a leading slash is required when using fully qualified class names.

So I started this PR to help others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,11 +147,11 @@ And then browsing to the host and port you specified
 // include your composer dependencies
 require_once 'vendor/autoload.php';
 
-$client = new Google\Client();
+$client = new \Google\Client();
 $client->setApplicationName("Client_Library_Examples");
 $client->setDeveloperKey("YOUR_APP_KEY");
 
-$service = new Google\Service\Books($client);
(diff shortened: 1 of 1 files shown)


bshaffer:
The samples are not set in a namespace, so omitting the `\` is appropriate. If your code file is namespaced, then the best practice would be to import the classes, not prefix them with a slash, e.g:

```php
use Google\Client;

$client = new Client();
```

xialeistudio:
> The samples are not set in a namespace, so omitting the `\` is appropriate. If your code file is namespaced, then the best practice would be to import the classes, not prefix them with a slash, e.g:
> 
> ```
> use Google\Client;
> 
> $client = new Client();
> ```

Understood, I believe this is an issue of document consistency, which may confuse users who are using the SDK for the first time. Therefore, I have submitted this PR to address the issue.

PR #2588, opened 2024-04-22
feat: Add the protected apiVersion property

Hectorhammett:
Add the protected `apiVersion` property to the `Resource` class to support the new api version header.


bshaffer (commented):
It's a small feature, but I'd still like to see a test for it!

Hectorhammett:
@bshaffer Added!

bshaffer (approved):

PR #2599, opened 2024-05-23
Add interfaces to be possible the mock of the client, batch and service

pedrocarmokununu:


bshaffer:
Hi @pedrocarmokununu !
Why do you need interfaces in order to mock these classes? They are not final, so you can mock them directly.

PR #2617, opened 2024-08-14
fix(docs): exclude aliases from reference documentation

bshaffer:


bshaffer:
This is no longer necessary because we switched to shared workflows in https://github.com/googleapis/google-api-php-client/pull/2618

PR #2625, opened 2024-09-03
Bump google-auth dep version

razvanphp:


bshaffer:
thank you @razvanphp !

Is there a specific reason why this needs to be bumped? I'm happy to do it, but I am curious because we do have `--prefer-lowest` tests which are passing, so if there's a compatibility error with a previous version, I would like to check why we aren't catching it.

razvanphp:
Ah, I see your point. The reason is so that we can use the features included in those new versions without risking not having them.

We have a dependancy on this repo only, not on the auth package. Maybe we can add this as explicit in our `composer.json` instead?

bshaffer:
closing because it's too difficult to get these tests to pass from a fork and with a non-google author, and since I updated the PR from main, now it requires two authors, which is a huge hassle.

PR #2643, opened 2024-12-03
chore: add php 8.4 to test matrix

bshaffer:


.github/workflows/tests.yml
```diff
@@ -11,12 +11,12 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ "8.0", "8.1", "8.2", "8.3" ]
+                php: [ "8.0", "8.1", "8.2", "8.3", "8.4" ]
```
Hectorhammett, line 14: The services repo does not depend on `Auth` but this one does, shouldn't we remove the "8.0" support?

Hectorhammett: Oh sorry, I miss-interpreted the conversation in our chat, I thought we were going to remove support here instead of my own PR.

Approved.

Hectorhammett (approved):

PR #2645, opened 2024-12-11
fix: add php 8.4 compatibility to latest 7.4 release

bshaffer:
fixes https://github.com/googleapis/google-api-php-client/issues/2640

Adds PHP 8.4 compatibility to the latest PHP 7.4 release (`2.16.0`). This will be tagged as `v2.16.1`


mikkamp:
Thanks for fixing this so quickly. I tested it with the following command:

```
composer require "google/apiclient:dev-2.16.x-php74 as 2.16.1"
```

And I can confirm that it's no longer showing the deprecation messages. So everything LGTM.

bshaffer:
This has been released! You can install it as

```
composer require google/apiclient:2.16.1
```

PR #2663, opened 2025-04-08
fix: convert Finder lazy iterator to array before deletion

ilyaplot:
Finder returns a live iterator that updates as files and directories are removed, causing some items to be skipped during deletion. This change converts the iterator to an array using `iterator_to_array()`, fixing the issue where not all files were deleted in one pass.

Closes https://github.com/googleapis/google-api-php-client-services/issues/6502


bshaffer (approved):

mustafacryptolife (commented):
.

mustafacryptolife (commented):
#+

PR #2670, opened 2025-05-07
fix: compatibility with new version of guzzle (guzzle > 5.3.0 & PHP >= 8)

emanuelgomez-s1:
This PR addresses a compatibility issue in the Google\Http\REST class related to legacy support for GuzzleHttp\Message\ResponseInterface, which was removed in Guzzle 6+. The existing code conditionally checks for this obsolete interface and attempts to convert responses, but this behavior is unnecessary (and potentially problematic) in modern environments using Guzzle >=6 and PHP 8.

The proposed change removes this outdated compatibility layer and ensures cleaner, future-proof code that aligns.

Thanks for your cooperation @macatapichon

### DIFF
--- a/.github/workflows/asset-release.yml
+++ b/.github/workflows/asset-release.yml
@@ -3,6 +3,7 @@ name: Add Release Assets
 on:
     release:
         types: [published]
+    workflow_dispatch:
 
 jobs:
     asset:


bshaffer:
@emanuelgomez-s1 
It seems to me that you have an issue in your autoloader if you're getting an error from the function `interface_exists`. There is no reason that line should cause a fatal PHP error

PR #2717, opened 2026-05-23
fix: cast generated header values to strings

GrahamCampbell:
Hi there. The Guzzle maintainer here again. I am working on the next major version of Guzzle with a focus on security hardening and reducing security-foot guns. As part of those changes, the next major version will disallow non-string header values. I will be triggering deprecations soon on the current major version series, which will start to impact this project, unless it is first modified to stop passing non-string header values. And thus, that is what this PR does.

---

FYI, there is a similar PR open on a sister repo at https://github.com/googleapis/google-cloud-php/pull/9208.

### DIFF
--- a/src/Http/MediaFileUpload.php
+++ b/src/Http/MediaFileUpload.php
@@ -297,9 +297,9 @@ class MediaFileUpload
         $body = $this->request->getBody();
         $headers = [
             'content-type' => 'application/json; charset=UTF-8',
-            'content-length' => $body->getSize(),
+            'content-length' => (string) $body->getSize(),
             'x-upload-content-type' => $this->mimeType,
-            'x-upload-content-length' => $this->size,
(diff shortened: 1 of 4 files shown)


bshaffer:
We need to update `symfony/dom-crawler` to version 5 or higher to resolve [the vulnerability](https://symfony.com/blog/cve-2026-45071-xxe-local-file-disclosure-in-domcrawler-addxmlcontent-via-validateonparse-true) which is failing the tests

GrahamCampbell:
I think that would be out of scope for this PR.

GrahamCampbell:
https://github.com/googleapis/google-api-php-client/pull/2718

bshaffer (approved):

Hectorhammett (approved):

SuhitNarayan:
hi help me

PR #2718, opened 2026-06-01
chore: upgrade Symfony crawler dependencies

GrahamCampbell:
I cross-checked the upstream changes and the usage in this codebase, and it looks like upgrading is as simple as just changing the version constraints.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,8 +18,8 @@
     },
     "require-dev": {
         "squizlabs/php_codesniffer": "^3.8",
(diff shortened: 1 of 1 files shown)


bshaffer (approved):

bshaffer:
@Hectorhammett the Generate Documentation job is failing because this is from a fork, and can be disregarded.

Hectorhammett (approved):

PR #2726, opened 2026-08-11
fix: match batch response parts to requests by Content-ID

sadiqk2:
Batch::parseResponse() paired each response part with a request by position ($requests[$i-1]), and silently ignored the $classes map that execute() builds keyed by Content-ID. The batch API does not guarantee that parts come back in the order they were sent — that is why each part carries a Content-ID — so a reordered response was decoded into the wrong model class (or into a raw Response), and alt=media detection used the wrong request.

This was a regression from the Guzzle 6 rewrite, which replaced the Content-ID lookup (`if (isset($classes[$key]))`) with positional matching and left $classes as an unused parameter.

Resolve both the originating request and the expected class from the part's Content-ID, falling back to the previous positional behaviour when the Content-ID is not one we sent.

### DIFF
--- a/src/Http/Batch.php
+++ b/src/Http/Batch.php
@@ -164,6 +164,13 @@ EOF;
             $responses = [];
             $requests = array_values($this->requests);
 
+            // Key the requests the same way the API keys the response parts, so
+            // a part can be matched to the request which produced it.
+            $requestsByContentId = [];
+            foreach ($this->requests as $requestKey => $batchedRequest) {
+                $requestsByContentId['response-' . $requestKey] = $batchedRequest;
+            }
+
             foreach ($parts as $i => $part) {
                 $part = trim($part);
                 if (!empty($part)) {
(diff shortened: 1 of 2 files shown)


src/Http/Batch.php
```diff
@@ -184,8 +191,18 @@ public function parseResponse(ResponseInterface $response, $classes = [])
                     // Need content id.
                     $key = $headers['content-id'];
 
+                    // The parts are not guaranteed to be returned in the order
+                    // they were sent, so resolve the request and the expected
+                    // class from the content id, and only fall back to the
+                    // position of the part when the content id is unknown.
+                    $request = $requestsByContentId[$key] ?? ($requests[$i-1] ?? null);
```
cy-yun: Since the null coalescing operator (`??`) can be chained safely, the parentheses around the fallback in your new code are not strictly necessary. You can simplify the assignment of `$request`.

```suggestion
                    // The parts are not guaranteed to be returned in the order
                    // they were sent, so resolve the request and the expected
                    // class from the content id, and only fall back to the
                    // position of the part when the content id is unknown.
                    $request = $requestsByContentId[$key] ?? $requests[$i-1] ?? null;
```

src/Http/Batch.php
```diff
@@ -184,8 +191,18 @@ public function parseResponse(ResponseInterface $response, $classes = [])
                     // Need content id.
                     $key = $headers['content-id'];
```
cy-yun: If the API returns a malformed or error part without a `Content-ID` header, accessing `$headers['content-id']` directly will emit an `E_WARNING` (Undefined array key). Furthermore, if `$key` evaluates to `null`, using it as an array key later (`$requestsByContentId[$key]` and `$classes[$key]`) will trigger a "Deprecated: Using null as an array offset" warning in PHP 8.1+. 

Using the null coalescing operator to fall back to an empty string safely avoids both issues.

```suggestion
                    // Need content id.
                    $key = $headers['content-id'] ?? '';
```

cy-yun (approved):
This is a great fix for the regression introduced by the Guzzle 6 rewrite. Re-associating the batched responses to their requests via the `Content-ID` header is the correct approach as per the Batch API specification, since parts are not guaranteed to maintain order. 

The tests thoroughly verify the logic. I've left a couple of minor inline suggestions to improve PHP 8.1+ compatibility and simplify the fallback logic, but overall the logic is sound!

bshaffer (changes requested):
> This was a regression from the Guzzle 6 rewrite

What guzzle 6 rewrite?

